PR #3, opened 2023-11-02
Add Serializer

bergundy:
Still not using it, just checking in the implementation to verify that the API is right.


bergundy:
I have a couple of thoughts:

1. I'm thinking of adding a `Header` type for convenience.

```go
type Header map[string]string

// Get is a case-insensitive key lookup from the header map.
func (h Header) Get(k string) string {
	return h[strings.ToLower(k)]
}
```

1. Since stream headers are transferred over HTTP, there's going to be collisions with the user provided request headers.
Only `Content-*` HTTP headers are used to construct a stream from an HTTP request / response.
I'm thinking that stream headers should not include the `Content-*` prefix, and that will be prepended by the client / server.

nexus/codec.go
```diff
@@ -0,0 +1,204 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"reflect"
+)
+
+// A Stream is a container for a header and a reader.
+// It is used to stream inputs and outputs in the various client and server APIs.
+type Stream struct {
```
cretz, line 14: Not the biggest fan of "stream" here as it implies long-lived endpoints w/ messages sent at differing times. Would prefer `Body` or `Content` or something.

nexus/codec.go
```diff
@@ -0,0 +1,204 @@
...
+func (s *LazyValue) Consume(v any) error {
+	defer s.Stream.Reader.Close()
+	return s.codec.Decode(s.Stream, v)
+}
+
+// Codec is used by the framework to serialize/deserialize input and output.
+// To customize serialization logic, implement this interface and provide your implementation to framework methods such
+// as [NewClient] and [NewHTTPHandler].
+// By default, the SDK supports serialization of JSONables, byte slices, and nils.
+type Codec interface {
```
cretz, line 47: A codec IMO is a bad term for value conversion, though an acceptable term for byte transform. I appreciate the separation between the two concepts that has grown in Temporal and I think we should maintain this separation instead of falling back to a combined data-converter approach.

nexus/codec.go
```diff
@@ -0,0 +1,204 @@
...
+		Reader: io.NopCloser(bytes.NewReader(body)),
+	}, nil
+}
+
+var _ Codec = jsonCodec{}
+
+type nilCodec struct{}
+
+func (nilCodec) Decode(s *Stream, v any) error {
+	if s.Header["content-length"] != "0" {
```
cretz, line 121: Be advised, not all HTTP clients set a content length for their requests. What happens if I make an empty Nexus call with no content type, no content length, etc? Does Nexus reject such a simple request?

nexus/codec.go
```diff
@@ -0,0 +1,204 @@
...
+		b, err := io.ReadAll(s.Reader)
+		if err != nil {
+			return err
+		}
+		*bPtr = b
+		return nil
+	}
+	// v is *any
+	rv := reflect.ValueOf(v).Elem()
+	if rv.Kind() == reflect.Interface {
```
cretz, line 172: I am a bit confused here on what you're trying to do. The code I think is saying every pointer to an interface you try to set as a byte array. You need to do exact equality check `Type()` with the reflect type of `any`, not any kind of interface.

But I am not convinced we should support `any` for this data type.

nexus/codec.go
```diff
@@ -0,0 +1,204 @@
...
+	return nil, errCodecIncompatible
+}
+
+var _ Codec = byteSliceCodec{}
+
+type compositeCodec struct {
+	codecChain
+}
+
+var defaultCodec Codec = compositeCodec{
```
cretz, line 202:
```suggestion
var DefaultCodec Codec = compositeCodec{
```

Should expose this IMO

bergundy: Let's leave this unexposed for now until it's deemed necessary.

nexus/codec.go
bergundy: `reflect.Interface` is `interface{}` (`any`).

I initially didn't support it but saw that Temporal does so I ported that logic.

It kinda makes sense that if you can populate an any, you would.

nexus/codec.go
bergundy: TBH, I'm not sure what to do here.

I guess I could try and read a byte of content and see if I get EOF and if it's not empty, mutate the reader to concatenate the bytes read with the rest of the content since the reader might not be seekable.
It just seems to complex and not worth it.

I also don't think I want to rely on content type here either.

nexus/codec.go
bergundy: Yeah, I'm changing to `Content`.

nexus/codec.go
bergundy: This isn't a combined thing. It's just for value to content and back.

We discussed having a mapper (`NewReader`, `NewWriter`) interface that's more like what Temporal considers a payload codec.

I've seen Codec used to describe a value encoder + decoder before (https://github.com/gcanti/io-ts/blob/master/Codec.md).

I'm open to renaming though, alternatives:

- `ContentConverter`
- `ValueConverter`
- `ValueCodec`
- `Marshaler` (very Go specific but even the Go one targets bytes and not a reader/writer)
- `Serializer`

I'm tempted to just call this `Codec` and later add a more specific interface that `Codec` will embed for value serialization.
That way when we add bytes to bytes conversion, it'll just be more methods that codec can implement:

```
type ValueCodec interface {
  FromValue(...)
  ToValue(...)
}

type IOMapper interface {
  NewWriter(...)
  NewReader(...)
}

type Codec interface {
  ValueCodec
  IOMapper
}
```

cretz: I don't think the word "codec" should be used for converting value to content and back. Some languages call this a serializer or a marshaler or a converter or something. If you don't like those for whatever reason, maybe we can find another that is in common use. I think codec is a bit of a strange term for this and I haven't really seen it used that way much in dev circles.

nexus/codec.go
cretz: We may be able to step back and ask why nil is needed/important as it's the absence of content. If for the JSON and byte slice converters we need to support absence of content we can. Can you show a situation where this is needed?

The way I am starting to think about it, only the people that actually read content should care about nil or not, and only once reading the content will that matter.

But yes, technically you could make a `bufio.NewReader` and invoke `Peek`.

nexus/codec.go
cretz: No, `reflect.Interface` is a value of the `Kind` enumerate which represents a _kind_ of a type, not the type. What most would do here is:

```
// store this somewhere global
var anyType := reflect.TypeOf((*any)(nil)).Elem()
rv.Type() == anyType
```

bergundy: Hmm... I thought I saw somewhere that `Interface` kind is `interface{}`. Thanks!

BTW: we should fix this in Temporal, it checks that the `Elem().Kind() == reflect.Interface` which is incorrect: https://github.com/temporalio/sdk-go/blob/0352634b45b1b81293c9bf3cec3bb033b82f841f/converter/byte_slice_payload_converter.go#L61

I think we should keep support for `[]byte -> any`, it'd be up to the user to cast.

nexus/codec.go
bergundy: I don't think we should mutate the content in the serializers (note using new terminology).

For now I think we should force setting a specific content type or content length 0.

nexus/codec.go
bergundy: I'm going with serializer, I don't love that it's unidirectional but I think it's common enough and not go specific that it's okay.

cretz: I like converter personally, but at this point it's whatever. I'm just glad that developer nomenclature is winning over individual concerns about unidirectional or whatever.

nexus/codec.go
cretz: I think we should support simple content-less HTTP calls. Maybe we shouldn't support optional JSON input. `[]byte` can just read 0 bytes and be fine, this whole thing is only about nullable JSON objects at this point and it's not right to make nullable JSON object be the reason that everyone must set some header even if they use `[]byte` or no parameter at all.

cretz (approved):
Not the biggest fan of the term "serializer" over "converter" and not a big fan that I can't give a simple HTTP request with no content headers at all (or does the serializer not get invoked when there are no parameters on the handler).

But otherwise LGTM.

bergundy:
> (or does the serializer not get invoked when there are no parameters on the handler)

We'll have a `NoValue` constant that we could potentially reference to avoid deserialization but we should support any nullable input.

It might be worth having the serializer only work on `[]byte` since there's not much value of supporting streaming there.
So `LazyValue.Consume` would first read the entire body.

PR #4, opened 2023-11-06
Operations

bergundy:
- Add top level `Operation` concept
- Change Nexus headers to accept single string value as a least common denominator for protocols we may want to support in the future
- Rename `Handler` interface to `ServiceHandler` to avoid encouraging its use of `OperationHandler`
- Add an `OperationDirectoryHandler` - likely will be used in most applications instead of implementing a `ServiceHandler` directly
- Unify client and server options structs
- Add a `HandlerErrorType` instead of relying on HTTP statuses

TODO:
- Fix/add all docstrings
- Fix README
- Add examples for using the operation APIs
- Consider abstracting HTTP in `completion.go`


nexus/operation.go
```diff
@@ -0,0 +1,220 @@
...
+func (d OperationDefinition[I, O]) Name() string {
+	return string(d)
+}
+
+// inferType implements Operation.
+func (OperationDefinition[I, O]) inferType(I, O) {} //nolint:unused
+
+var _ Operation[any, any] = OperationDefinition[any, any]("")
+
+type UntypedOperationHandler interface {
```
bergundy, line 32:
```suggestion
type RegisterableOperationHandler interface {
```

nexus/operation.go
```diff
@@ -0,0 +1,220 @@
...
+	Operations []UntypedOperationHandler
+}
+
+type OperationDirectoryHandler struct {
+	UnimplementedServiceHandler
+
+	operations map[string]UntypedOperationHandler
+}
+
+func NewOperationDirectoryHandler(options OperationDirectoryHandlerOptions) (*OperationDirectoryHandler, error) {
```
bergundy, line 87:
```
r := NewOperationRegistry()
r.Register(...operations)
NewOperationRegistryHandler(r, options)
```

```
r := NewOperationRegistry()
r.Register(...operations)
r.NewServiceHandler(options)
```

```
opts := OperationDirectoryHandlerOptions{}
append(opts.Operations, ...)
```

```
r := OperationRegistry{}
r.Register(...operations)
r.NewServiceHandler(options) (ServiceHandler, error)
```

bergundy: We'll go with the last one

bergundy (commented):
```
ServiceHandler -> Handler
OperationDirectoryHandler -> Registry.NewHandler()
OperationHandler -> Operation
NewOperationDefinition (or Reference)
```

cretz (approved):
This is good enough to get moving with (any Go API designs we can revisit later)

PR #6, opened 2024-01-05
Integrate Serializers for completion

bergundy:
This is making up for some tech debt where I introduced Serializers to the SDK and didn't apply it for the completion code.
I need this now for integrating into the server repo.

### DIFF
--- a/nexus/completion.go
+++ b/nexus/completion.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"strconv"
 )
 
 // NewCompletionHTTPRequest creates an HTTP request deliver an operation completion to a given URL.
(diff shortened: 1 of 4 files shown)


nexus/completion_test.go
```diff
@@ -40,10 +41,11 @@ func TestSuccessfulCompletion(t *testing.T) {
 	ctx, callbackURL, teardown := setupForCompletion(t, &successfulCompletionHandler{})
 	defer teardown()
 
-	request, err := NewCompletionHTTPRequest(ctx, callbackURL, &OperationCompletionSuccessful{
-		Header: http.Header{"foo": []string{"bar"}},
-		Body:   bytes.NewReader([]byte("success")),
-	})
+	completion, err := NewOperationCompletionSuccessful("content", OperationCompletionSuccesfulOptions{})
```
MichaelSnowden, line 44: I think we ought to have a test case where the serializer is set.

nexus/completion.go
```diff
@@ -115,11 +140,14 @@ type CompletionHandlerOptions struct {
 	// A stuctured logging handler.
 	// Defaults to slog.Default().
 	Logger *slog.Logger
+	// A [Serializer] to customize handler serialization behavior.
+	// By default the handler handles, JSONables, byte slices, and nil.
+	Serializer Serializer
```
MichaelSnowden, line 145: I'd be interested in knowing what the default serializer does when the handler returns something not in this list.

MichaelSnowden (approved):

nexus/completion_test.go
bergundy: I was too lazy to add it but I'll do it before merging.

nexus/completion.go
bergundy: It's very much like the default payload converter in temporal SDKs.
It will fail if it can't serialize to JSON.

PR #9, opened 2024-04-08
Propagate context deadline to request timeout header

pdoerner:
Added handling for propagating context deadlines
* For client (outgoing) requests: the context deadline is set on the HTTP request header `Request-Timeout`
* For for server (incoming) requests: if set, the `Request-Timeout` HTTP request header is applied to the context returned by the HTTP request

### DIFF
--- a/nexus/cancel_test.go
+++ b/nexus/cancel_test.go
@@ -95,8 +95,9 @@ func TestCancel_ContextDeadlinePropagated(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestCancel_RequestTimeoutHeaderPropagated(t *testing.T) {
+func TestCancel_RequestTimeoutHeaderOverridesContextDeadline(t *testing.T) {
 	timeout := 100 * time.Millisecond
+	// relies on ctx returned here having default testTimeout set greater than expected timeout
 	ctx, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{expectedTimeout: timeout})
 	defer teardown()
 
--- a/nexus/cancel_test.go
+++ b/nexus/cancel_test.go
@@ -108,12 +108,11 @@ func TestCancel_RequestTimeoutHeaderOverridesContextDeadline(t *testing.T) {
 }
 
 func TestCancel_TimeoutNotPropagated(t *testing.T) {
-	ctx, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{})
+	_, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{})
 	defer teardown()
 
-	ctx = context.Background()
 	handle, err := client.NewHandle("foo", "timeout")
 	require.NoError(t, err)
-	err = handle.Cancel(ctx, CancelOperationOptions{})
+	err = handle.Cancel(context.Background(), CancelOperationOptions{})
 	require.NoError(t, err)
 }
--- a/nexus/cancel_test.go
+++ b/nexus/cancel_test.go
@@ -3,6 +3,7 @@ package nexus
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
 )
@@ -53,6 +54,64 @@ func TestCancel_HandleFromClient(t *testing.T) {
 	defer teardown()
 
 	handle, err := client.NewHandle("f/o/o", "a/sync")
+	require.NoError(t, err)
+	err = handle.Cancel(ctx, CancelOperationOptions{})
+	require.NoError(t, err)
+}
+
+type echoTimeoutAsyncWithCancelHandler struct {
+	expectedTimeout time.Duration
+	UnimplementedHandler
+}
+
+func (h *echoTimeoutAsyncWithCancelHandler) StartOperation(ctx context.Context, operation string, input *LazyValue, options StartOperationOptions) (HandlerStartOperationResult[any], error) {
+	return &HandlerStartOperationResultAsync{
+		OperationID: "timeout",
+	}, nil
+}
+
+func (h *echoTimeoutAsyncWithCancelHandler) CancelOperation(ctx context.Context, operation, operationID string, options CancelOperationOptions) error {
+	deadline, set := ctx.Deadline()
+	if h.expectedTimeout > 0 && !set {
+		return HandlerErrorf(HandlerErrorTypeBadRequest, "expected operation to have timeout set but context has no deadline")
+	}
+	if h.expectedTimeout <= 0 && set {
+		return HandlerErrorf(HandlerErrorTypeBadRequest, "expected operation to have no timeout but context has deadline set")
+	}
+	timeout := time.Until(deadline)
+	if timeout > h.expectedTimeout {
+		return HandlerErrorf(HandlerErrorTypeBadRequest, "operation has timeout (%s) greater than expected (%s)", timeout.String(), h.expectedTimeout.String())
+	}
+	return nil
+}
+
+func TestCancel_ContextDeadlinePropagated(t *testing.T) {
+	ctx, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{expectedTimeout: testTimeout})
+	defer teardown()
+
+	handle, err := client.NewHandle("foo", "timeout")
+	require.NoError(t, err)
+	err = handle.Cancel(ctx, CancelOperationOptions{})
+	require.NoError(t, err)
+}
+
+func TestCancel_RequestTimeoutHeaderPropagated(t *testing.T) {
+	timeout := 100 * time.Millisecond
+	ctx, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{expectedTimeout: timeout})
+	defer teardown()
+
+	handle, err := client.NewHandle("foo", "timeout")
+	require.NoError(t, err)
+	err = handle.Cancel(ctx, CancelOperationOptions{Header: Header{headerRequestTimeout: timeout.String()}})
+	require.NoError(t, err)
+}
+
+func TestCancel_TimeoutNotPropagated(t *testing.T) {
+	ctx, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{})
+	defer teardown()
+
+	ctx = context.Background()
+	handle, err := client.NewHandle("foo", "timeout")
 	require.NoError(t, err)
 	err = handle.Cancel(ctx, CancelOperationOptions{})
 	require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


nexus/server.go
```diff
@@ -311,8 +326,21 @@ func (h *httpHandler) getOperationResult(writer http.ResponseWriter, request *ht
 	}
 	options := GetOperationResultOptions{Header: httpHeaderToNexusHeader(request.Header)}
 
-	waitStr := request.URL.Query().Get(queryWait)
+	// If both Request-Timeout http header and wait query string are set, the minimum will be used as the context timeout.
```
bergundy, line 329: This should be the minimum between `h.options.GetResultTimeout` and the parsed header value.

nexus/server.go
```diff
@@ -288,7 +288,22 @@ func (h *httpHandler) startOperation(writer http.ResponseWriter, request *http.R
...
+		timeoutDuration, err := time.ParseDuration(timeoutStr)
+		if err != nil {
+			h.logger.Warn("invalid request timeout header", "timeout", timeoutStr)
+			h.writeFailure(writer, HandlerErrorf(HandlerErrorTypeBadRequest, "invalid request timeout header"))
+			return
+		}
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(request.Context(), timeoutDuration)
+		defer cancel()
+	}
```
bergundy, line 304: looks like you can refactor this to a method.

bergundy (commented):
Overall LGTM.
Go ahead and add tests.

nexus/get_info_test.go
```diff
@@ -64,3 +65,123 @@ func TestGetInfoHandleFromClientNoHeader(t *testing.T) {
...
 	require.Equal(t, OperationStateCanceled, info.State)
 }
+
+type asyncWithInfoTimeoutHandler struct {
+	UnimplementedHandler
+}
+
+func (h *asyncWithInfoTimeoutHandler) StartOperation(ctx context.Context, operation string, input *LazyValue, options StartOperationOptions) (HandlerStartOperationResult[any], error) {
+	return &HandlerStartOperationResultAsync{
+		OperationID: "needs /URL/ escaping",
```
bergundy, line 75: nit: you can use a simpler operation ID since you're not testing escaping.

nexus/get_info_test.go
```diff
@@ -64,3 +65,123 @@ func TestGetInfoHandleFromClientNoHeader(t *testing.T) {
...
+		},
+		{
+			name:         "time_out: set on context and header",
+			timeout:      1 * time.Millisecond,
+			setOnHeader:  true,
+			setOnContext: true,
+			validator: func(t *testing.T, handle *OperationHandle[*LazyValue], info *OperationInfo, err error) {
+				require.ErrorContains(t, err, "context deadline exceeded")
+			},
+		},
```
bergundy, line 130: I would test that the header overwrites the context deadline and not go beyond that.

bergundy: You'd want contextTimeout and headerTimeout to be set separately to test this.

nexus/get_info_test.go
```diff
@@ -64,3 +65,123 @@ func TestGetInfoHandleFromClientNoHeader(t *testing.T) {
...
+			name:         "success: set on context and header",
+			timeout:      5 * time.Second,
+			setOnHeader:  true,
+			setOnContext: false,
+			validator: func(t *testing.T, handle *OperationHandle[*LazyValue], info *OperationInfo, err error) {
+				require.NoError(t, err)
+				require.Equal(t, handle.ID, info.ID)
+				require.Equal(t, OperationStateCanceled, info.State)
+			},
+		},
```
bergundy, line 163: You don't need this coverage, check that if you don't set a context deadline from the client, the server doesn't set a context deadline either.

You can achieve this by sending a header to the client to tell it what to expect.

nexus/start_test.go
```diff
@@ -250,3 +252,127 @@ func TestUnsuccessful(t *testing.T) {
...
+	if ctx.Err() != nil && errors.Is(ctx.Err(), context.DeadlineExceeded) {
+		return nil, HandlerErrorf(HandlerErrorTypeDownstreamTimeout, "handler exceeded request timeout of %s", options.Header.Get(headerRequestTimeout))
+	}
+
+	return &HandlerStartOperationResultSync[any]{
+		Value: []byte(options.Header.Get("Request-Timeout")),
+	}, nil
+}
+
+func TestRequestTimeout(t *testing.T) {
```
bergundy, line 272: You don't need all of the cases here, see my comments in `get_info_test.go`.

nexus/server.go
```diff
@@ -386,14 +412,51 @@ func (h *httpHandler) cancelOperation(writer http.ResponseWriter, request *http.
...
 	options := CancelOperationOptions{Header: httpHeaderToNexusHeader(request.Header)}
 
-	if err := h.options.Handler.CancelOperation(request.Context(), operation, operationID, options); err != nil {
+	ctx, cancel, ok := h.contextWithTimeoutFromHTTPRequest(writer, request)
+	if !ok {
+		return
+	}
+	defer cancel()
+
+	if err := h.options.Handler.CancelOperation(ctx, operation, operationID, options); err != nil {
```
bergundy, line 421: CancelOperation doesn't have test coverage.

nexus/get_result_test.go
```diff
@@ -124,6 +124,19 @@ func TestWaitResult_DeadlineExceeded(t *testing.T) {
...
+func TestWaitResult_RequestTimeout(t *testing.T) {
+	ctx, client, teardown := setup(t, &asyncWithResultHandler{timesToBlock: 1000})
+	defer teardown()
+
+	result, err := client.StartOperation(ctx, "foo", nil, StartOperationOptions{})
+	require.NoError(t, err)
+	handle := result.Pending
+	require.NotNil(t, handle)
+
+	_, err = handle.GetResult(ctx, GetOperationResultOptions{Wait: time.Second, Header: Header{headerRequestTimeout: "200ms"}})
```
bergundy, line 136: Let's not test the header, use context instead.

nexus/get_result_test.go
```diff
@@ -124,6 +124,19 @@ func TestWaitResult_DeadlineExceeded(t *testing.T) {
...
+	require.ErrorIs(t, err, ErrOperationStillRunning)
```
bergundy, line 137: Let's test that the timeout is actually propagated.

nexus/get_info_test.go
```diff
@@ -64,3 +65,123 @@ func TestGetInfoHandleFromClientNoHeader(t *testing.T) {
...
+		validator    func(t *testing.T, handle *OperationHandle[*LazyValue], info *OperationInfo, err error)
+	}
+	cases := []testcase{
+		{
+			name:         "time_out: set on context",
+			timeout:      1 * time.Millisecond,
+			setOnHeader:  false,
+			setOnContext: true,
+			validator: func(t *testing.T, handle *OperationHandle[*LazyValue], info *OperationInfo, err error) {
+				require.ErrorContains(t, err, "context deadline exceeded")
```
bergundy, line 110: Use `require.ErrorIs(t, err, context.DeadlineExceeded)`

nexus/get_info_test.go
bergundy: You only want 3 tests for every API:

1. context deadline is propagated to the server
2. request header overwrites header from context deadline
3. no context deadline is set on server if there's no context deadline in client (with a special case for GetResult with `wait`).

cretz:
This is only semi-related to this PR...

Can we make sure any of our Go HTTP servers and clients have overarching timeouts set on them? They shouldn't be this suggested request-timeout header probably, but they need to be some max value to prevent Slowloris type of attacks.

nexus/cancel_test.go
```diff
@@ -57,3 +58,61 @@ func TestCancel_HandleFromClient(t *testing.T) {
...
+	ctx, client, teardown := setup(t, &echoTimeoutAsyncWithCancelHandler{expectedTimeout: testTimeout})
+	defer teardown()
+
+	handle, err := client.NewHandle("foo", "timeout")
+	require.NoError(t, err)
+	err = handle.Cancel(ctx, CancelOperationOptions{})
+	require.NoError(t, err)
+}
+
+func TestCancel_RequestTimeoutHeaderPropagated(t *testing.T) {
```
bergundy, line 98: Please also put a shorter timeout on the context and see that the header overrides the context deadline in this test.

```suggestion
func TestCancel_RequestTimeoutHeaderOverridesContextDeadline(t *testing.T) {
```

nexus/get_info_test.go
```diff
@@ -64,3 +65,65 @@ func TestGetInfoHandleFromClientNoHeader(t *testing.T) {
...
+	if h.expectedTimeout > 0 && !set {
+		return nil, HandlerErrorf(HandlerErrorTypeBadRequest, "expected operation to have timeout set but context has no deadline")
+	}
+	if h.expectedTimeout <= 0 && set {
+		return nil, HandlerErrorf(HandlerErrorTypeBadRequest, "expected operation to have no timeout but context has deadline set")
+	}
+	timeout := time.Until(deadline)
+	if timeout > h.expectedTimeout {
+		return nil, HandlerErrorf(HandlerErrorTypeBadRequest, "operation has timeout (%s) greater than expected (%s)", timeout.String(), h.expectedTimeout.String())
+	}
```
bergundy, line 91: nit: can you put this in a helper method?

nexus/get_info_test.go
```diff
@@ -64,3 +65,65 @@ func TestGetInfoHandleFromClientNoHeader(t *testing.T) {
...
+	ctx, client, teardown := setup(t, &asyncWithInfoTimeoutHandler{expectedTimeout: testTimeout})
+	defer teardown()
+
+	handle, err := client.NewHandle("foo", "timeout")
+	require.NoError(t, err)
+	_, err = handle.GetInfo(ctx, GetOperationInfoOptions{})
+	require.NoError(t, err)
+}
+
+func TestGetInfo_RequestTimeoutHeaderPropagated(t *testing.T) {
```
bergundy, line 109: Same as above please.

bergundy: There's no doubt it'll be propagated, but we want to test that it overrides.

nexus/start_test.go
```diff
@@ -250,3 +251,70 @@ func TestUnsuccessful(t *testing.T) {
...
+
+	deadline, _ := ctx.Deadline()
+	initialTimeout := time.Until(deadline)
+	result, err := client.StartOperation(ctx, "foo", nil, StartOperationOptions{})
+
+	require.NoError(t, err)
+	requireTimeoutPropagated(t, result, initialTimeout)
+}
+
+func TestStart_RequestTimeoutHeaderPropagated(t *testing.T) {
```
bergundy, line 283: Same as above please.

bergundy (approved):
I'm approving but there are still a few small comments to address.

PR #14, opened 2024-06-17
Remove downstream error type and status code

pdoerner:
Removing the downstream error handler error type and response code.
Changed downstream timeout to code 520.


nexus/api.go
```diff
@@ -40,7 +40,6 @@ const (
 	statusOperationRunning = http.StatusPreconditionFailed
 	// HTTP status code for failed operation responses.
 	statusOperationFailed   = http.StatusFailedDependency
-	StatusDownstreamError   = 520
 	StatusDownstreamTimeout = 521
```
bergundy, line 43: See my comment for using 520 for timeout in the API PR.

bergundy (approved):

PR #15, opened 2024-07-22
Remove gorilla mux dependency

bergundy:
Also bump version to 0.0.9.

## Why?

We want to minimize dependencies for our SDKs.

### DIFF
--- a/nexus/server.go
+++ b/nexus/server.go
@@ -12,9 +12,8 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
-
-	"github.com/gorilla/mux"
 )
 
 // An HandlerStartOperationResult is the return type from the [Handler] StartOperation and [Operation] Start methods. It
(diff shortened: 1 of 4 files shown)


nexus/server.go
```diff
@@ -488,6 +428,66 @@ type HandlerOptions struct {
...
+		return
+	}
+	operation, err := url.PathUnescape(parts[2])
+	if err != nil {
+		h.writeFailure(writer, HandlerErrorf(HandlerErrorTypeBadRequest, "failed to parse URL path"))
+		return
+	}
+	var operationID string
+	if len(parts) > 3 {
+		var err error
```
cretz, line 450:
```suggestion
```

Not needed (unimportant change though)

nexus/server.go
```diff
@@ -488,6 +428,66 @@ type HandlerOptions struct {
...
+		}
+		h.startOperation(service, operation, writer, request)
+	case 4: // /{service}/{operation}/{operation_id}
+		if request.Method != "GET" {
+			h.writeFailure(writer, HandlerErrorf(HandlerErrorTypeBadRequest, "invalid request method: expected GET, got %q", request.Method))
+			return
+		}
+		h.getOperationInfo(service, operation, operationID, writer, request)
+	case 5:
+		switch parts[4] {
```
cretz, line 472: Can you add a `default` clause here failing?

bergundy: I extracted the fallback outside of the `default:` clause.

cretz (approved):

PR #17, opened 2024-08-20
Add Link to StartOperationOptions

rodrigozhou:
Encode/Decode Nexus links to/from HTTP headers.


nexus/api.go
```diff
@@ -22,6 +23,7 @@ const (
 	headerOperationState = "Nexus-Operation-State"
 	headerOperationID    = "Nexus-Operation-Id"
 	headerRequestID      = "Nexus-Request-Id"
+	headerLinks          = "Nexus-Links"
```
bergundy, line 26:
```suggestion
	headerLink          = "Nexus-Link"
```

nexus/api.go
```diff
@@ -176,3 +203,28 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
 }
+
+type Link struct {
+	Data []byte `json:"data"`
+	Type string `json:"type"`
+}
+
+var linkBase64Encoding = base64.StdEncoding
+
+func EncodeLink(link *Link) (string, error) {
```
bergundy, line 214:
```suggestion
func EncodeLink(link Link) (string, error) {
```

nexus/api.go
```diff
@@ -176,3 +203,28 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
 	httpHeader.Set(HeaderRequestTimeout, time.Until(deadline).String())
 	return httpHeader
 }
+
+type Link struct {
+	Data []byte `json:"data"`
+	Type string `json:"type"`
+}
+
+var linkBase64Encoding = base64.StdEncoding
```
bergundy, line 212: Redunant

nexus/api.go
```diff
@@ -176,3 +203,28 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+
+func EncodeLink(link *Link) (string, error) {
+	linkJson, err := json.Marshal(link)
+	if err != nil {
+		return "", err
+	}
+	return linkBase64Encoding.EncodeToString(linkJson), nil
+}
+
+func DecodeLink(encodedLink string) (*Link, error) {
```
bergundy, line 222:
```suggestion
func DecodeLink(encodedLink string) (Link, error) {
```

nexus/api.go
bergundy, line 214: Add a docstring please (here and below), this is an exported method.

nexus/client.go
```diff
@@ -196,6 +203,9 @@ func (c *Client) StartOperation(ctx context.Context, operation string, input any
 	addCallbackHeaderToHTTPHeader(options.CallbackHeader, request.Header)
 	addContextTimeoutToHTTPHeader(ctx, request.Header)
 	addNexusHeaderToHTTPHeader(options.Header, request.Header)
+	if err := addLinksToHTTPHeader(options.Links, request.Header); err != nil {
+		return nil, err
```
bergundy, line 207:
```suggestion
		return nil, fmt.Errorf("failed to add links to header: %w", err)
```

bergundy: Put the nexus header last, it intentionally overrides all of the other headers.

nexus/client.go
```diff
@@ -229,12 +239,21 @@ func (c *Client) StartOperation(ctx context.Context, operation string, input any
 		if info.State != OperationStateRunning {
 			return nil, newUnexpectedResponseError(fmt.Sprintf("invalid operation state in response info: %q", info.State), response, body)
 		}
+		links, err := getLinksFromHeader(response.Header)
+		if err != nil {
+			return nil, newUnexpectedResponseError(
```
bergundy, line 244: Maybe wrap with `fmt.Errorf("%w: %w", newUnexpectedResponseError(...), err)` to avoid losing the original error.

nexus/server.go
```diff
@@ -264,11 +273,17 @@ func (h *baseHTTPHandler) writeFailure(writer http.ResponseWriter, err error) {
 }
 
 func (h *httpHandler) startOperation(service, operation string, writer http.ResponseWriter, request *http.Request) {
+	links, err := getLinksFromHeader(request.Header)
+	if err != nil {
+		h.writeFailure(writer, err)
```
bergundy, line 278: Make sure this returns a bad request.

nexus/options.go
```diff
@@ -26,6 +26,8 @@ type StartOperationOptions struct {
 	// Request ID that may be used by the server handler to dedupe a start request.
 	// By default a v4 UUID will be generated by the client.
 	RequestID string
+	// Links contain information about the caller event that made the Nexus call.
```
bergundy, line 29: I'm wondering if we need to support multiple links in this direction, I was thinking a slice was only required in the response. I will think about it and bring it up for discussion.

bergundy, line 29:
```suggestion
	// Links contain arbitrary caller information, e.g. a URL or an arbitrary structure serialized to bytes.
```

nexus/api.go
```diff
@@ -176,3 +203,28 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
 	httpHeader.Set(HeaderRequestTimeout, time.Until(deadline).String())
 	return httpHeader
 }
+
+type Link struct {
+	Data []byte `json:"data"`
+	Type string `json:"type"`
```
bergundy, line 209: Docstrings.

bergundy (commented):
Overall direction LGTM.
I know this is intentional but noting that this is missing docs and tests.

nexus/api.go
rodrigozhou: The goal was to make sure that `EncodeLink` and `DecodeLink` are using the same base64 encoding object (in this case, `base64.StdEncoding`).

nexus/api.go
rodrigozhou: It could have multiple values. Is it convention to have header names always in singular?

nexus/client.go
rodrigozhou: I wonder if we should return error at all since this is the case the response returned status CREATED, and it's just the links header that was unable to be decoded.

nexus/api.go
rodrigozhou: Hm, actually, I don't think we need to export these methods.

nexus/api.go
bergundy: I don't think there's convention around this.
I wanted to use the singular form because each header key value pair represents a single link in our use case but I'd be fine with going with plural too.

cretz: It is fairly common for HTTP headers that accept multiple values (they can be provided comma delimited or as separate lines) to be singular. For instance it's `Accept-Language` not `Accept-Languages`.

rodrigozhou: I'm using the same header for the HTTP response, which we know it can contain multiple links (although it's only one for our first use case).
I'll change to singular.

nexus/api.go
bergundy: Yeah, I don't think that justifies this variable though. But if you feel strongly about it, keep it.

rodrigozhou: Removed.

nexus/api.go
```diff
@@ -22,6 +25,7 @@ const (
 	headerOperationState = "Nexus-Operation-State"
 	headerOperationID    = "Nexus-Operation-Id"
 	headerRequestID      = "Nexus-Request-Id"
+	headerLinks          = "Nexus-Link"
```
bergundy, line 28:
```suggestion
	headerLink          = "Nexus-Link"
```

nexus/api.go
```diff
@@ -176,3 +199,65 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
 	httpHeader.Set(HeaderRequestTimeout, time.Until(deadline).String())
 	return httpHeader
 }
+
+type Link struct {
+	// URL encoded information about the link.
+	URL string `json:"url"`
+	// Type can describe an actual data type for decoding the URL.
+	Type string `json:"type"`
```
bergundy, line 207: Don't need JSON tags anymore.

nexus/api.go
```diff
@@ -176,3 +199,65 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+}
+
+func encodeLink(link Link) string {
+	return fmt.Sprintf("<%s>; Type=%q", link.URL, link.Type)
```
bergundy, line 211:
```suggestion
	return fmt.Sprintf("%s; type=%q", link.URL, link.Type)
```

nexus/api.go
```diff
@@ -176,3 +199,65 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+	consumeUntilAndDiscard := func(buf []byte, c byte) (bool, []byte, []byte) {
+		idx := bytes.IndexByte(buf, c)
+		if idx == -1 {
+			return false, buf, nil
+		}
+		return true, bytes.TrimSpace(buf[:idx]), bytes.TrimSpace(buf[idx+1:])
+	}
+
+	var link Link
+	buf := []byte(encodedLink)
```
bergundy, line 224: Why convert to `[]byte`?

nexus/api.go
```diff
@@ -145,6 +149,25 @@ func addCallbackHeaderToHTTPHeader(nexusHeader Header, httpHeader http.Header) h
...
+	for _, link := range links {
+		httpHeader.Add(headerLinks, encodeLink(link))
+	}
+	return nil
+}
+
+func getLinksFromHeader(httpHeader http.Header) ([]Link, error) {
+	var links []Link
+	for _, encodedLink := range httpHeader.Values(headerLinks) {
+		link, err := decodeLink(encodedLink)
```
bergundy, line 162: I would double check that Go automatically splits up headers that have multiple values (e.g. you won't see commas here).

rodrigozhou: From what I can tell, Go only accepts multi-value header by specifying multiple times the same header name.

nexus/api.go
rodrigozhou: Re-wrote the function assuming that certain chars are forbidden. It simplified the implementation.

nexus/api.go
```diff
@@ -145,6 +146,25 @@ func addCallbackHeaderToHTTPHeader(nexusHeader Header, httpHeader http.Header) h
 	return httpHeader
 }
 
+func addLinksToHTTPHeader(links []Link, httpHeader http.Header) error {
```
cretz, line 149: No need to model the non-exported function as returning an error if it never does (refactoring is safe to add it later if needed). Having said that, this should return an error because `encodeLink` should return an error on invalid link.

nexus/api.go
```diff
@@ -145,6 +146,25 @@ func addCallbackHeaderToHTTPHeader(nexusHeader Header, httpHeader http.Header) h
...
+	for _, link := range links {
+		httpHeader.Add(headerLink, encodeLink(link))
+	}
+	return nil
+}
+
+func getLinksFromHeader(httpHeader http.Header) ([]Link, error) {
+	var links []Link
+	for _, encodedLink := range strings.Split(strings.Join(httpHeader.Values(headerLink), ","), ",") {
```
cretz, line 158:
```suggestion
	for _, encodedLink := range httpHeader.Values(headerLink) {
```

I would expect Go is already doing the comma-delimited separation for you. Can you confirm?

bergundy: https://github.com/nexus-rpc/sdk-go/pull/17/files#r1728048912

nexus/api.go
bergundy: I checked, Go will split up the header values.

nexus/api.go
bergundy: I verified using `curl ... -H "foo: a, b"` and this line of code in an http handler: `fmt.Println(r.Header.Values("foo")[0])`.

I got `a, b`

nexus/server.go
```diff
@@ -48,6 +49,14 @@ func (r *HandlerStartOperationResultAsync) applyToHTTPResponse(writer http.Respo
 		return
 	}
 
+	if err := addLinksToHTTPHeader(r.Links, writer.Header()); err != nil {
+		handler.logger.Error("failed to serialize links info", "error", err)
```
bergundy, line 53:
```suggestion
		handler.logger.Error("failed to serialize links into header", "error", err)
```

nexus/server.go
```diff
@@ -264,11 +273,17 @@ func (h *baseHTTPHandler) writeFailure(writer http.ResponseWriter, err error) {
 }
 
 func (h *httpHandler) startOperation(service, operation string, writer http.ResponseWriter, request *http.Request) {
+	links, err := getLinksFromHeader(request.Header)
+	if err != nil {
+		h.writeFailure(writer, HandlerErrorf(HandlerErrorTypeBadRequest, "invalid links header"))
```
bergundy, line 278:
```suggestion
		h.writeFailure(writer, HandlerErrorf(HandlerErrorTypeBadRequest, "invalid %q header", headerLink))
```

nexus/options.go
```diff
@@ -26,6 +26,8 @@ type StartOperationOptions struct {
 	// Request ID that may be used by the server handler to dedupe a start request.
 	// By default a v4 UUID will be generated by the client.
 	RequestID string
+	// Links contain arbitrary caller information, e.g. a URL or an arbitrary structure serialized to bytes.
```
bergundy, line 29:
```suggestion
	// Links contain arbitrary caller information. Handers may use these links as metadata on resources associated with an operation.
```

nexus/client.go
```diff
@@ -268,6 +291,8 @@ type ExecuteOperationOptions struct {
 	// Request ID that may be used by the server handler to dedupe this start request.
 	// By default a v4 UUID will be generated by the client.
 	RequestID string
+	// Links contain information about the caller event that made the Nexus call.
```
bergundy, line 294: Copy my suggestion for the docstring from `StartOperationOptions`.

nexus/api_test.go
```diff
@@ -74,3 +74,144 @@ func TestFailure_JSONMarshalling(t *testing.T) {
...
+		name   string
+		input  string
+		output Link
+		errMsg string
+	}
+
+	cases := []testcase{
+		{
+			name:  "valid",
+			input: `<temporal:///namespaces/test-ns/workflows/test-wf-id/test-run-id/history?event_id=1&event_type=2>; Type="temporal.api.common.v1.Link.WorkflowEvent"`,
```
bergundy, line 115: Don't reference temporal in generic nexus code please.

nexus/api_test.go
```diff
@@ -74,3 +74,144 @@ func TestFailure_JSONMarshalling(t *testing.T) {
 		})
 	}
 }
+
+func TestEncodeLink(t *testing.T) {
```
bergundy, line 78: Please add a case for when the URL is invalid and prove that you generate a valid header if the URL requires quoting and such.

nexus/api_test.go
```diff
@@ -74,3 +74,144 @@ func TestFailure_JSONMarshalling(t *testing.T) {
...
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			output := encodeLink(tc.input)
+			require.Equal(t, tc.expected, output)
+		})
+	}
+}
+
+func TestDecodeLink(t *testing.T) {
```
bergundy, line 104: Please test multiple links in a single header.

bergundy: (e.g. comma delimited)

nexus/api.go
```diff
@@ -176,3 +196,53 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+		}
+		key := strings.TrimSpace(kv[0])
+		val := strings.TrimSpace(kv[1])
+		if strings.HasPrefix(val, `"`) != strings.HasSuffix(val, `"`) {
+			return link, fmt.Errorf("failed to parse link header value: %q", encodedLink)
+		}
+		if strings.HasPrefix(val, `"`) {
+			val = val[1 : len(val)-1]
+		}
+		if key == "Type" {
```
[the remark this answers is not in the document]
bergundy: +1. I'm sure I requested this yesterday, can't find my comment now.

nexus/api.go
cretz: :+1:, can ignore me here.

nexus/api.go
```diff
@@ -176,3 +196,53 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
 	httpHeader.Set(HeaderRequestTimeout, time.Until(deadline).String())
 	return httpHeader
 }
+
+type Link struct {
+	// URL encoded information about the link
+	// It cannot contain the following chars: ',', '';', '='.
+	URL string
```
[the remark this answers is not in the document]
rodrigozhou: We need to manually validate anyway since we allow workflow ID to contain any char. For the URL, the workflow ID will have to be percent-encoded. Otherwise, decoding the Link header will be a pain.

nexus/api.go
```diff
@@ -176,3 +196,53 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+	url := strings.TrimSpace(parts[0])
+	if !strings.HasPrefix(url, "<") || !strings.HasSuffix(url, ">") {
+		return link, fmt.Errorf("failed to parse link header value: %q", encodedLink)
+	}
+	link.URL = url[1 : len(url)-1]
+
+	for _, part := range parts[1:] {
+		if len(part) == 0 {
+			continue
+		}
```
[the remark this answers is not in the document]
rodrigozhou: When there's a trailing semi-colon. Eg: `<url>; param=value;`

nexus/api.go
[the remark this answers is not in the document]
rodrigozhou: Thought about this, and chose to let it overwrite, so the last value set is the value stored. I'm ok either way.

nexus/api.go
rodrigozhou: Yes, I did the same, and Go does not auto-split the values.

nexus/api.go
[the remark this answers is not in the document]
cretz: Not following this exactly. Whatever you do with workflow ID extraction from the URL is unrelated to this general-purpose, Temporal-agnostic library. I was just suggesting that data type because it is guaranteed a valid URL if in that data type. But if we want validated string instead that's fine.

nexus/api.go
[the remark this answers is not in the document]
cretz: I am not sure that should be considered ok

nexus/api.go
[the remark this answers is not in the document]
cretz: > You mean require the values to be double-quoted?

No, require the key be present, and arguably require that it be `type` for now

> What about if we allow it to be case-insensitive?

We should not allow this IMO

nexus/api.go
```diff
@@ -176,3 +203,82 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+	// Valid chars: alphanumeric, '_', '.', '/'
+	Type string
+}
+
+const linkTypeKey = "type"
+
+func encodeLink(link Link) (string, error) {
+	u, err := url.Parse(link.URL)
+	if err != nil {
+		return "", fmt.Errorf("failed to encode link: url not valid: %s", link.URL)
```
cretz, line 224:
```suggestion
		return "", fmt.Errorf("failed to encode link: url %s not valid: %w", link.URL, err)
```

nexus/api.go
```diff
@@ -176,3 +203,82 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+		}
+	}
+	_, err = url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return "", fmt.Errorf("failed to encode link: query not percent-encoded: %s", link.URL)
+	}
+
+	validSymbols := []rune("_./")
+	for _, c := range link.Type {
+		if !unicode.IsLetter(c) && !unicode.IsDigit(c) && !slices.Contains(validSymbols, c) {
```
cretz, line 239: Unicode letters and digits aren't the same as alnum (I think they allow non-ascii chars). Consider:

```suggestion
		if (c < 'a' || c > 'z') && (c < 'A' || c > 'Z') && (c < '0' || c > '9') && c != '_' && c != '.' && c != '/' {
```

Or something like that (didn't test)

nexus/api.go
```diff
@@ -176,3 +203,82 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+		}
+		key := strings.TrimSpace(kv[0])
+		val := strings.TrimSpace(kv[1])
+		if strings.HasPrefix(val, `"`) != strings.HasSuffix(val, `"`) {
+			return link, fmt.Errorf("failed to parse link header value: %s", encodedLink)
+		}
+		if strings.HasPrefix(val, `"`) {
+			val = val[1 : len(val)-1]
+		}
+		if strings.ToLower(key) == linkTypeKey {
```
cretz, line 278:
```suggestion
		if key == linkTypeKey {
```

I don't think this needs to be case insensitive. Also, arguably you should validate the value for type too before returning a potentially bad link that will encoding. Consider making the URL and type validators utilities you call on encode and decode.

nexus/api.go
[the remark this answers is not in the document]
rodrigozhou: I'm gonna require the `type` key to be present, and ignore any additional keys for backwards/forward compatibility.

nexus/api.go
```diff
@@ -176,3 +203,82 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+			link.Type = val
+		}
+	}
+
+	return link, nil
```
cretz, line 283: I don't think we should support links without `type`.

nexus/api.go
```diff
@@ -176,3 +203,82 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+			return "", fmt.Errorf("failed to encode link: path not percent-encoded: %s", link.URL)
+		}
+	}
+	_, err = url.ParseQuery(u.RawQuery)
+	if err != nil {
+		return "", fmt.Errorf("failed to encode link: query not percent-encoded: %s", link.URL)
+	}
+
+	validSymbols := []rune("_./")
+	for _, c := range link.Type {
```
cretz, line 238: What if type is empty string? I don't think we should allow it.

nexus/api.go
```diff
@@ -176,3 +201,107 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+	return fmt.Sprintf(`<%s>; %s="%s"`, link.URL, linkTypeKey, link.Type), nil
+}
+
+func decodeLink(encodedLink string) (Link, error) {
+	var link Link
+	parts := strings.Split(encodedLink, ";")
+
+	u := strings.TrimSpace(parts[0])
+	if !strings.HasPrefix(u, "<") || !strings.HasSuffix(u, ">") {
+		return link, fmt.Errorf("failed to parse link header value: %s", encodedLink)
```
bergundy, line 235: Maybe add some context?

```suggestion
		return link, fmt.Errorf("failed to parse link header value: %q, expected URL to be enclosed between < and >", encodedLink)
```

nexus/api.go
```diff
@@ -176,3 +201,107 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+	URL string
+	// Type can describe an actual data type for decoding the URL.
+	// Valid chars: alphanumeric, '_', '.', '/'
+	Type string
+}
+
+const linkTypeKey = "type"
+
+func encodeLink(link Link) (string, error) {
+	if err := validateLinkURL(link.URL); err != nil {
```
bergundy, line 220: It would probably be easier for users to provide a URL here and let us serialize it than worry about them properly encoding the values.

nexus/api.go
```diff
@@ -176,3 +201,107 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+func decodeLink(encodedLink string) (Link, error) {
+	var link Link
+	parts := strings.Split(encodedLink, ";")
+
+	u := strings.TrimSpace(parts[0])
+	if !strings.HasPrefix(u, "<") || !strings.HasSuffix(u, ">") {
+		return link, fmt.Errorf("failed to parse link header value: %s", encodedLink)
+	}
+	u = u[1 : len(u)-1]
+	if err := validateLinkURL(u); err != nil {
```
bergundy, line 238: Just url.Parse is good enough here.

nexus/api.go
rodrigozhou: Well, technically, the URL needs to be correctly encoded to be a valid URL, doesn't it? Eg: `a/b/c` vs `a/b%2Fc`. Here, `%2F` is the encoded value of `/`.

nexus/api.go
rodrigozhou: It's not. `url.Parse("/foo<>;/bar") does not fail, although the path is not properly encoded.

nexus/client.go
bergundy: You resolved without changing.

nexus/api.go
rodrigozhou: Changed the `URL` type to `url.URL`.

nexus/api.go
```diff
@@ -176,3 +201,128 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+			linkTypeKey,
+			encodedLink,
+		)
+	}
+
+	return link, nil
+}
+
+func validateLinkURL(value *url.URL) error {
+	if value.String() == "" {
```
cretz, line 308: I think this panics if URL is nil

nexus/client.go
rodrigozhou: It's changed.

nexus/api.go
```diff
@@ -176,3 +201,128 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header)
...
+	if err != nil {
+		return link, fmt.Errorf("failed to parse link header: invalid url: %s", urlStr)
+	}
+	if err := validateLinkURL(u); err != nil {
+		return link, fmt.Errorf("failed to parse link header: %w", err)
+	}
+	link.URL = u
+
+	params := strings.Split(encodedLink[urlEnd+1:], ";")
+	// must contain at least one semi-colon, and first param must be empty since
```
cretz, line 261: > first param must be empty

This doesn't appear to be validated, but not a big deal

cretz (approved):

PR #20, opened 2024-09-13
Limit duration format according to spec

bergundy:
We used to accept anything `time.ParseDuration` would parse which is too broad and violates the spec.
Also ensured that we use a custom `formatDuration` implementation to ensure compatibility.

### DIFF
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -11,6 +11,8 @@ import (
 	"mime"
 	"net/http"
 	"net/url"
+	"regexp"
+	"strconv"
 	"strings"
 	"time"
 )
@@ -202,7 +204,7 @@ func addContextTimeoutToHTTPHeader(ctx context.Context, httpHeader http.Header) 
 	if !ok {
 		return httpHeader
 	}
-	httpHeader.Set(HeaderRequestTimeout, time.Until(deadline).String())
+	httpHeader.Set(HeaderRequestTimeout, formatDuration(time.Until(deadline)))
 	return httpHeader
 }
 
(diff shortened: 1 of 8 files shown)


nexus/api.go
```diff
@@ -333,3 +335,32 @@ func validateLinkType(value string) error {
 	}
 	return nil
 }
+
+var durationRegexp = regexp.MustCompile("^(\\d+)(ms|s|m|h)$")
```
cretz, line 339: I think floats should be supported. Spec says "Format of this parameter is number + unit" and I took "number" to mean float as did the Java PR.

Quinn-With-Two-Ns: Yes , the server does generate floats

Quinn-With-Two-Ns: Also `h` is not supported by the spec?

> Format of this header value is number + unit, where unit can be `ms` for milliseconds, `s` for seconds, and `m` for
minutes.

bergundy: Hmmm... if the server generates floats we'll need to support parsing it for compatibility but I'd stop generating floats on the server ASAP and use ints instead.
Definitely can get rid of `h`, thanks for pointing that out, I should've checked, I was looking at @cretz's comment in the Java PR (`"s", "m", or "h"`).

Quinn-With-Two-Ns: Is there a reason we didn't do ISO 8601 durations?

bergundy: Just felt like overkill and we wanted a format that can be easily typed in a curl command.

Quinn-With-Two-Ns (approved):

PR #21, opened 2024-10-04
Add Operation InputType and OutputType helper methods

bergundy:
Useful helper methods, we need this in the Temporal Go SDK to validate inputs when starting an operation from a workflow.

### DIFF
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -20,6 +20,8 @@ type NoValue *struct{}
 // available.
 type OperationReference[I, O any] interface {
 	Name() string
+	// IsInputCompatible returns a boolean indicating whether the given input is assignable to the generic I type.
+	IsInputAssignable(any) bool
 	// A type inference helper for implementations of this interface.
 	inferType(I, O)
 }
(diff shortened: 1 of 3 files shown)


nexus/operation.go
```diff
@@ -36,6 +38,13 @@ func (r operationReference[I, O]) Name() string {
 	return string(r)
 }
 
+func (r operationReference[I, O]) IsInputAssignable(i any) bool {
```
cretz, line 41: Not sure this needs to be on every operation reference (nor does `IsOutputAssignable`). If you had a `InputType() reflect.Type` that could maybe make a bit more sense, but anyone can obtain the generic of the operation reference via reflection. IMO having a helper for Go reflection is unnecessary for this public API

Can just update the Temporal side to use reflection if it needs to check whether something is assignable, same as it would for any generic API it would need to do a similar check for.

bergundy: I don't think it's possible to get the generic input type in Go. I tried this before and ended up blocked on https://github.com/golang/go/issues/54393.

Maybe there's another way you're aware of that I'm not.

cretz: I see, my mistake. I think `InputType() reflect.Type` and `OutputType() reflect.Type` would be better general-purpose methods to add.

bergundy: That works for me.

nexus/operation.go
```diff
@@ -36,6 +40,16 @@ func (r operationReference[I, O]) Name() string {
 	return string(r)
 }
 
+func (operationReference[I, O]) InputType() reflect.Type {
+	var zero [0]I
+	return reflect.TypeOf(zero).Elem()
```
cretz, line 45:
```suggestion
	return reflect.TypeOf((*I)(nil)).Elem()
```

This may also work (untested), but what you have is fine/cheap too, no need to change.

cretz (approved):

PR #22, opened 2024-10-10
Expose API to get operations from the Service

rodrigozhou:
Add `Service.GetOperation` to be able to get an operation by its name.

### DIFF
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -166,6 +166,12 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
 	return nil
 }
 
+// GetOperation returns the operation by its name. The bool value indicates if the operation was found.
+func (s *Service) GetOperation(name string) (RegisterableOperation, bool) {
+	h, ok := s.operations[name]
+	return h, ok
+}
+
 // A ServiceRegistry registers services and constructs a [Handler] that dispatches operations requests to those services.
 type ServiceRegistry struct {
 	services map[string]*Service
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -166,10 +166,9 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
 	return nil
 }
 
-// GetOperation returns the operation by its name. The bool value indicates if the operation was found.
-func (s *Service) GetOperation(name string) (RegisterableOperation, bool) {
-	h, ok := s.operations[name]
-	return h, ok
+// Operation returns an operation by name or nil if not found.
+func (s *Service) Operation(name string) RegisterableOperation {
+	return s.operations[name]
 }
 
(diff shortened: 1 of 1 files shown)


nexus/operation.go
```diff
@@ -166,6 +166,12 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
 	return nil
 }
 
+// GetOperation returns the operation by its name. The bool value indicates if the operation was found.
+func (s *Service) GetOperation(name string) (RegisterableOperation, bool) {
```
bergundy, line 170:
```suggestion
// Operation returns an operation by name and a boolean indicating if the operation was found.
func (s *Service) Operation(name string) (RegisterableOperation, bool) {
```

cretz, line 170: Is this needed? We haven't needed this on SDK workers today. We haven't traditionally added getters for things the user sets. Notice how there isn't a `Client.GetOptions` or a `LazyValue.GetSerializer` or any of that.

I didn't see the use case in the description, but I am curious why a user might not prefer a collection of all the operations (so it can also get the count or iterate through them, e.g. to get all operations that start with a certain value).

bergundy: This is used here: https://github.com/temporalio/sdk-go/pull/1666/files#diff-0fe12a5bacd5440b4c9a31cf9f1f5fe0aa50826d976254054e5df783a84c8435R3197.

Seems reasonable to expose this.

cretz: :+1: Might as well expose all operations. Also, should consider just returning nil if it is not found, no need for a second result.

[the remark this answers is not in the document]
bergundy: Ah true, it's possible since this returns an interface.

cretz: > Looks like cobra expose both Flag and Flags methods

A bad example I think (compared to the many dozens that can show otherwise)

I am not a big fan of increasing API surface for internal use only or, when we do override that concern, increasing it in a limited myopic way only thinking about our limited/narrow use.

bergundy (approved):

nexus/operation.go
bergundy: You would still want a method to extract a single operation instead of having to construct a slice every time to list all operations and then search that list.

I don't mind if we add another method to list all operations if you feel like it would be more complete, no strong opinion here. I already approved the PR.

nexus/operation.go
```diff
@@ -166,6 +166,11 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
 	return nil
 }
 
+// Operation returns an operation by name or nil if not found.
+func (s *Service) Operation(name string) RegisterableOperation {
+	return s.operations[name]
```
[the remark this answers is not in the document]
rodrigozhou: I'll handle this in a separate PR

PR #23, opened 2024-10-10
Add Operation-Timeout header

pdoerner:
Added a new exported constant for the `Operation-Timeout` header.

Added logic to set the context deadline and outgoing `Request-Timeout` header to the minimum of the incoming `Request-Timeout` and `Operation-Timeout` if they are set.

### DIFF
--- a/nexus/server.go
+++ b/nexus/server.go
@@ -293,7 +293,7 @@ func (h *httpHandler) startOperation(service, operation string, writer http.Resp
 		},
 	}
 
-	ctx, cancel, ok := h.contextWithTimeoutFromHTTPRequest(writer, request)
+	ctx, cancel, ok := h.contextWithTimeoutFromHTTPRequest(writer, request, HeaderOperationTimeout)
 	if !ok {
 		return
 	}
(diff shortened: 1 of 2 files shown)


nexus/server.go
```diff
@@ -395,30 +395,44 @@ func (h *httpHandler) cancelOperation(service, operation, operationID string, wr
...
-// the Request-Timeout header, if set.
-func (h *httpHandler) contextWithTimeoutFromHTTPRequest(writer http.ResponseWriter, request *http.Request) (context.Context, context.CancelFunc, bool) {
-	requestTimeout, ok := h.parseRequestTimeoutHeader(writer, request)
+// contextWithTimeoutFromHTTPRequest extracts the context from the HTTP request and applies the minimum timeout of the
+// timeouts indicated by the Request-Timeout header and any additional specified headers, if set.
+func (h *httpHandler) contextWithTimeoutFromHTTPRequest(writer http.ResponseWriter, request *http.Request, headerNames ...string) (context.Context, context.CancelFunc, bool) {
+	requestTimeout, ok := h.parseTimeoutHeader(writer, request, HeaderRequestTimeout)
 	if !ok {
 		return nil, nil, false
 	}
```
cretz, line 421: What if I only set an operation timeout? Could just:

```suggestion
	headerNames = append(headerNames, HeaderRequestTimeout)
```

(only acceptable to mutate the trailing param because this is a private function and you know caller isn't reusing header name slice)

pdoerner: Only setting an operation timeout is valid. The `parseTimeoutHeader` will only return `!ok` if there is a parsing error. It will return `0, true` if unset.

About modifying the trailing param: I thought about a couple different options but didn't love any of them. I settled on this because I thought it made it more explicit that `Request-Timeout` is always checked. But I am not committed to this approach if others feel strongly.

cretz: You're good, I misunderstood what absence of a request timeout did here. No need to change anything. (may be worth a test or two though)

cretz (approved):

pdoerner:
Discussed with @bergundy offline and decided this extra logic is unnecessary.

PR #25, opened 2024-10-15
Add a lock to nexus.Service to protect access to operations map

rodrigozhou:
Add a lock to `nexus.Service` to protect access to `operations` map, and thus making it thread safe.

### DIFF
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -133,7 +133,7 @@ func (h *syncOperation[I, O]) Start(ctx context.Context, input I, options StartO
 type Service struct {
 	Name string
 
-	// Lock to protect operations map in case new services are added while workflows are running.
+	// Lock to protect the operations map in case new operations are added while workflows are running.
 	lock       sync.RWMutex
 	operations map[string]RegisterableOperation
 }
@@ -167,7 +167,7 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
 	return nil
 }
 
-// addOperation adds the operation if not found. Returns boolean indicating if the operation was added.
+// addOperation adds the operation if not found. Returns a boolean indicating if the operation was added.
 func (s *Service) addOperation(op RegisterableOperation) bool {
 	s.lock.Lock()
 	defer s.lock.Unlock()
@@ -185,8 +185,8 @@ func (s *Service) Operation(name string) RegisterableOperation {
 	return s.operations[name]
 }
 
-// NumOperations returns the number of operations registered.
-func (s *Service) NumOperations() int {
+// numOperations returns the number of operations registered.
+func (s *Service) numOperations() int {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 	return len(s.operations)
@@ -230,7 +230,7 @@ func (r *ServiceRegistry) NewHandler() (Handler, error) {
 		return nil, errors.New("must register at least one service")
 	}
 	for _, service := range r.services {
-		if service.NumOperations() == 0 {
+		if service.numOperations() == 0 {
 			return nil, fmt.Errorf("service %q has no operations registered", service.Name)
 		}
 	}
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"sync"
 )
 
 // NoValue is a marker type for an operations that do not accept any input or return a value (nil).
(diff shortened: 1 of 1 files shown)


nexus/operation.go
```diff
@@ -132,6 +133,8 @@ func (h *syncOperation[I, O]) Start(ctx context.Context, input I, options StartO
 type Service struct {
 	Name string
 
+	// Lock to protect operations map in case new services are added while workflows are running.
```
bergundy, line 136:
```suggestion
	// Lock to protect the operations map in case new services are added while workflows are running.
```

nexus/operation.go
```diff
@@ -166,11 +167,31 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
 	return nil
 }
 
+// addOperation adds the operation if not found. Returns boolean indicating if the operation was added.
```
bergundy, line 170:
```suggestion
// addOperation adds the operation if not found. Returns a boolean indicating if the operation was added.
```

nexus/operation.go
```diff
@@ -166,11 +167,31 @@ func (s *Service) Register(operations ...RegisterableOperation) error {
...
 	return s.operations[name]
 }
 
+// NumOperations returns the number of operations registered.
+func (s *Service) NumOperations() int {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	return len(s.operations)
+}
+
```
bergundy, line 194: Don't expose this please.

bergundy (commented):
I can't think of a way to seal the service after creation, which would have been desirable. But I think it's better to provide the safety than what we had before.

nexus/operation.go
```diff
@@ -132,6 +133,8 @@ func (h *syncOperation[I, O]) Start(ctx context.Context, input I, options StartO
 type Service struct {
 	Name string
 
+	// Lock to protect operations map in case new services are added while workflows are running.
+	lock       sync.RWMutex
```
cretz, line 137: Does this really need to be thread safe? Or can we copy the operations at registration time. Or are we intentionally supporting registering operations after a service is already registered?

cretz: > It doesn't need to be thread safe but I worry that users will not read our docstrings and assume it is

That's not really how Go works. Everyone is supposed to assume not safe unless docs say otherwise. They will have issues with every other Go library including the standard library with those assumptions.

> We could disable late registration by adding a Seal() method to Service that is automatically called when the service is registered.

Not sure it should be public, but sure anything to prevent post-service-registration operation registration works for me

bergundy: It would need to be public for the Temporal SDK to call it.

Let's go ahead and do that then.
I think the bool doesn't need to be atomic, since it can be set under the lock and never gets unset. Let's make this thread safe, doesn't really hurt anything.

cretz: > It would need to be public for the Temporal SDK to call it.

Ah, I figured they'd use the registry that's in here. Then I would not suggest `Seal()`, I would suggest `Clone()` that returns a copied instance.

bergundy: Hmm.. I think there's value in a `Seal()` API. It'd give a clear error if a user tries to register operations post handler creation. `Clone()` would be silent. Another (somewhat negligible) advantage to `Seal()` is that operation lookup doesn't require a lock or atomic variable if we do add thread safety to the registry.

cretz: It's not a very common Go approach (nor is `Freeze` or terms like that). Usually you mutate something and you pass it. You might not expect the thing you pass it to to also mutate (by calling `Seal`).

bergundy: So maybe let's just keep this not thread safe and let users deal with the implications of that if it's a more common Go approach.
Could be good to stress that the fact that it's not thread safe also means that users should not try to register operations after the service has been registered.

cretz: Works for me. And really, the "Seal" thing works for me too if you'd like to actually prevent post-servce-registration operation registration, though by-copy/clone type of approaches are more common.

cretz (approved):
Approving because I don't have blocking opinions (it's ok to make them thread safe, there's no harm in it). I also support telling people not to register operations after service is registered, and I also support disallowing operation registration after service registration in clone/copy form or "seal"/"freeze" form.

bergundy:
I'm also okay with what this PR does. I can settle with simply putting a comment saying we discourage late registration (need to come up with good rationale for that though if we mention it).

@rodrigozhou feel free to merge as is or incorporate any of the suggestions.

bergundy:
Closing due to lack of interest.

PR #26, opened 2024-11-15
Add OperationID, StartTime and StartLinks to CompletionRequest

pdoerner:
Added `OperationID`, `StartTime` and `StartLinks` fields to `CompletionRequest`. This information is necessary to fabricate a start event if the completion is received before the response to the StartOperation request.

Bumps SDK version to `0.0.12`

### DIFF
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -22,18 +22,19 @@ const version = "v0.0.12"
 
 const (
 	// Nexus specific headers.
-	headerOperationState = "Nexus-Operation-State"
-	headerRequestID      = "Nexus-Request-Id"
-	headerLink           = "Nexus-Link"
+	headerOperationState     = "Nexus-Operation-State"
+	headerRequestID          = "Nexus-Request-Id"
+	headerLink               = "Nexus-Link"
+	headerOperationStartTime = "Nexus-Operation-Start-Time"
+	// HeaderOperationID is the unique ID returned by the StartOperation response for async operations.
+	// Must be set on callback headers to support completing operations before the start response is received.
+	HeaderOperationID = "Nexus-Operation-Id"
 
 	// HeaderRequestTimeout is the total time to complete a Nexus HTTP request.
 	HeaderRequestTimeout = "Request-Timeout"
 	// HeaderOperationTimeout is the total time to complete a Nexus operation.
 	// Unlike HeaderRequestTimeout, this applies to the whole operation, not just a single HTTP request.
 	HeaderOperationTimeout = "Operation-Timeout"
-	// HeaderOperationID is the unique ID returned by the StartOperation response for async operations.
-	// Must be set on callback headers to support completing operations before the start response is received.
-	HeaderOperationID = "Nexus-Operation-Id"
 )
 
 const contentTypeJSON = "application/json"
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -18,12 +18,11 @@ import (
 )
 
 // Package version.
-const version = "v0.0.11"
+const version = "v0.0.12"
 
 const (
 	// Nexus specific headers.
 	headerOperationState = "Nexus-Operation-State"
-	headerOperationID    = "Nexus-Operation-Id"
 	headerRequestID      = "Nexus-Request-Id"
 	headerLink           = "Nexus-Link"
 
@@ -32,6 +31,9 @@ const (
 	// HeaderOperationTimeout is the total time to complete a Nexus operation.
 	// Unlike HeaderRequestTimeout, this applies to the whole operation, not just a single HTTP request.
 	HeaderOperationTimeout = "Operation-Timeout"
+	// HeaderOperationID is the unique ID returned by the StartOperation response for async operations.
+	// Must be set on callback headers to support completing operations before the start response is received.
+	HeaderOperationID = "Nexus-Operation-Id"
 )
 
 const contentTypeJSON = "application/json"
--- a/nexus/completion.go
+++ b/nexus/completion.go
@@ -38,9 +38,11 @@ type OperationCompletionSuccessful struct {
 	// Body to send in the completion HTTP request.
 	// If it implements `io.Closer` it will automatically be closed by the client.
 	Body io.Reader
-	// StartTime is the time the operation started. Used when a completion request is received before a started response.
+	// OperationID is the unique ID for this operation. Used when a completion callback is received before a started response.
+	OperationID string
+	// StartTime is the time the operation started. Used when a completion callback is received before a started response.
 	StartTime time.Time
-	// StartLinks are the links attached to the started response. Used when a completion request is received before a started response.
+	// StartLinks are the links attached to the started response. Used when a completion callback is received before a started response.
 	StartLinks []Link
 }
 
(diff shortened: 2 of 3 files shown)


nexus/api.go
```diff
@@ -18,12 +18,11 @@ import (
 )
 
 // Package version.
-const version = "v0.0.11"
+const version = "v0.0.12"
```
cretz, line 21: What is this version number? I don't see it in https://github.com/nexus-rpc/api. Do we need to start tagging version?

cretz (approved):

nexus/completion.go
```diff
@@ -69,8 +80,10 @@ func NewOperationCompletionSuccessful(result any, options OperationCompletionSuc
 		header := http.Header{"Content-Length": []string{strconv.Itoa(len(content.Data))}}
 
 		return &OperationCompletionSuccessful{
-			Header: addContentHeaderToHTTPHeader(content.Header, header),
-			Body:   bytes.NewReader(content.Data),
+			Header:     addContentHeaderToHTTPHeader(content.Header, header),
```
bergundy, line 83: We've typically let user provided headers override any well known headers that the SDK sets. Let's preserve this behavior here.

nexus/api.go
cretz: I see. I wonder if we want a separate PR for that just before release, but if we know we're just about to release this small change as another version, ok.

nexus/completion.go
```diff
@@ -37,21 +38,31 @@ type OperationCompletionSuccessful struct {
...
 
 // OperationCompletionSuccessfulOptions are options for [NewOperationCompletionSuccessful].
 type OperationCompletionSuccessfulOptions struct {
 	// Optional serializer for the result. Defaults to the SDK's default Serializer, which handles JSONables, byte
 	// slices and nils.
 	Serializer Serializer
+	// StartTime is the time the operation started. Used when a completion request is received before a started response.
+	StartTime time.Time
+	// StartLinks are the links attached to the started response. Used when a completion request is received before a started response.
+	StartLinks []Link
```
bergundy, line 55: Let's add OperationID here too even if Temporal won't use it for completeness.

nexus/completion.go
```diff
@@ -95,6 +115,10 @@ type OperationCompletionUnsuccessful struct {
 	Header http.Header
 	// State of the operation, should be failed or canceled.
 	State OperationState
+	// StartTime is the time the operation started. Used when a completion request is received before a started response.
+	StartTime time.Time
+	// StartLinks are the links attached to the started response. Used when a completion request is received before a started response.
+	StartLinks []Link
```
bergundy, line 121: Same here, please add OperationID.

nexus/completion_test.go
```diff
@@ -41,8 +49,20 @@ func TestSuccessfulCompletion(t *testing.T) {
...
+				Host:     "example.com",
+				Path:     "/path/to/something",
+				RawQuery: "param=value",
+			},
+			Type: "url",
+		}},
+	})
 	completion.Header.Add("foo", "bar")
+	completion.Header.Add(HeaderOperationID, "test-operation-id")
+	completion.Header.Add(headerOperationStartTime, time.Now().Format(http.TimeFormat))
```
bergundy, line 65: Provide the start time via options?

nexus/completion_test.go
```diff
@@ -55,15 +75,25 @@ func TestSuccessfulCompletion(t *testing.T) {
 	require.Equal(t, http.StatusOK, response.StatusCode)
 }
 
-func TestSuccessfulCompletion_CustomSerializr(t *testing.T) {
+func TestSuccessfulCompletion_CustomSerializer(t *testing.T) {
```
bergundy, line 78: Thanks!

nexus/completion_test.go
```diff
@@ -92,6 +122,12 @@ func (h *failureExpectingCompletionHandler) CompleteOperation(ctx context.Contex
 	if completion.HTTPRequest.Header.Get("foo") != "bar" {
 		return HandlerErrorf(HandlerErrorTypeBadRequest, "invalid 'foo' header: %q", completion.HTTPRequest.Header.Get("foo"))
 	}
+	if completion.OperationID != "test-operation-id" {
+		return HandlerErrorf(HandlerErrorTypeBadRequest, "invalid %q header: %q", HeaderOperationID, completion.HTTPRequest.Header.Get(HeaderOperationID))
+	}
+	if len(completion.StartLinks) == 0 {
+		return HandlerErrorf(HandlerErrorTypeBadRequest, "expected StartLinks to be set on CompletionRequest")
+	}
```
bergundy, line 130: nit: you're already checking this in the other handler. I'd just leave it at that.

bergundy (commented):
Added some small comments. Nothing major. Feel free to merge after addressing.

nexus/completion.go
pdoerner: I updated the logic in the `applyToHTTPRequest` methods. LMK if that looks good to you.

nexus/completion_test.go
```diff
@@ -92,6 +122,9 @@ func (h *failureExpectingCompletionHandler) CompleteOperation(ctx context.Contex
 	if completion.HTTPRequest.Header.Get("foo") != "bar" {
 		return HandlerErrorf(HandlerErrorTypeBadRequest, "invalid 'foo' header: %q", completion.HTTPRequest.Header.Get("foo"))
 	}
+	if completion.OperationID != "test-operation-id" {
+		return HandlerErrorf(HandlerErrorTypeBadRequest, "invalid %q header: %q", HeaderOperationID, completion.HTTPRequest.Header.Get(HeaderOperationID))
+	}
```
bergundy, line 127: Hmm.. I think I may have mislead you in the previous review. We should test that the links are delivered here, it's a different path in the caller for unsuccessful operations.

nexus/completion.go
```diff
@@ -95,6 +126,12 @@ type OperationCompletionUnsuccessful struct {
 	Header http.Header
 	// State of the operation, should be failed or canceled.
 	State OperationState
+	// OperationID is the unique ID for this operation. Used when a completion callback is received before a started response.
+	OperationID string
+	// StartTime is the time the operation started. Used when a completion callback is received before a started response.
+	StartTime time.Time
+	// StartLinks are the links attached to the started response. Used when a completion callback is received before a started response.
+	StartLinks []Link
```
bergundy, line 134: I'm wondering if it's worth just calling these `Links`. Because there's no difference between a start link and a complete link.

Also, I'd improve the docstring here:

```suggestion
	// StartLinks are used to link back to the operation when a completion callback is received before a started response.
	StartLinks []Link
```

bergundy (approved):
Approved with two small comments.

nexus/completion.go
pdoerner: I was thinking users may want to differentiate between links that should be attached to a fabricated start event and links that should be attached to the complete event. But maybe that is a Temporal-specific implementation detail?

bergundy: This should've been called `Links` IMHO. I will rename and submit it with other breaking changes I have planned.

PR #27, opened 2024-11-25
Make all header keys lower case and add Set utility

bergundy:
This change copies what Go `http.Header` does but uses lower case as the canonical form.
Before this change, calling `nexus.Header{nexus.HeaderRequestTimeout: "1s"}` would produce an invalid header value.

### DIFF
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -22,16 +22,16 @@ const version = "v0.0.11"
 
 const (
 	// Nexus specific headers.
-	headerOperationState = "Nexus-Operation-State"
-	headerOperationID    = "Nexus-Operation-Id"
-	headerRequestID      = "Nexus-Request-Id"
-	headerLink           = "Nexus-Link"
+	headerOperationState = "nexus-operation-state"
+	headerOperationID    = "nexus-operation-id"
+	headerRequestID      = "nexus-request-id"
+	headerLink           = "nexus-link"
 
 	// HeaderRequestTimeout is the total time to complete a Nexus HTTP request.
-	HeaderRequestTimeout = "Request-Timeout"
+	HeaderRequestTimeout = "request-timeout"
 	// HeaderOperationTimeout is the total time to complete a Nexus operation.
 	// Unlike HeaderRequestTimeout, this applies to the whole operation, not just a single HTTP request.
-	HeaderOperationTimeout = "Operation-Timeout"
+	HeaderOperationTimeout = "operation-timeout"
 )
 
 const contentTypeJSON = "application/json"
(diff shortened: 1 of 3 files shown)


nexus/api_test.go
```diff
@@ -99,7 +100,7 @@ func TestAddLinksToHeader(t *testing.T) {
 				Type: "url",
 			}},
 			output: http.Header{
-				headerLink: []string{
+				textproto.CanonicalMIMEHeaderKey(headerLink): []string{
```
pdoerner, line 103: Why use this library over `http.CanonicalHeaderKey`?

pdoerner (approved):

nexus/api_test.go
bergundy: I missed this, they’re essentially the same function but it’s cleaner to use the function exported in http. Thanks!

nexus/api.go
```diff
@@ -119,13 +119,19 @@ func isMediaTypeOctetStream(contentType string) bool {
 
 // Header is a mapping of string to string.
 // It is used throughout the framework to transmit metadata.
+// The keys should be in lower case form.
 type Header map[string]string
```
Quinn-With-Two-Ns, line 123: Shouldn't this be a list to match HTTP?

Quinn-With-Two-Ns (changes requested):
This change would break clients if deployed on the server. I think we need to fix the clients before changing the strip representation.

nexus/api.go
bergundy: No, we deliberately made Nexus headers take only one value.

nexus/api.go
```diff
@@ -22,16 +22,16 @@ const version = "v0.0.11"
 
 const (
 	// Nexus specific headers.
-	headerOperationState = "Nexus-Operation-State"
-	headerOperationID    = "Nexus-Operation-Id"
-	headerRequestID      = "Nexus-Request-Id"
-	headerLink           = "Nexus-Link"
+	headerOperationState = "nexus-operation-state"
```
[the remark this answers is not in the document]
bergundy: A design decision that was made when we started this project.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: Is that called our here https://github.com/nexus-rpc/api/blob/main/SPEC.md? All of the documentation for headers is in the previous format and I can't find any reference to lowercase

Quinn-With-Two-Ns:
This line in the Java SDK would break https://github.com/temporalio/sdk-java/blob/a2dd36942b27c8be6d0445c9fe5b33f6240a2050/temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java#L99 as currently `Request-Timeout` from the server is sent as "Request-Timeout" and the SDK does not do a case insensitive comparison

bergundy:
> This line in the Java SDK would break https://github.com/temporalio/sdk-java/blob/a2dd36942b27c8be6d0445c9fe5b33f6240a2050/temporal-sdk/src/main/java/io/temporal/internal/nexus/NexusTaskHandlerImpl.java#L99 as currently `Request-Timeout` from the server is sent as "Request-Timeout" and the SDK does not do a case insensitive comparison

I see, I missed this in the Java review. I will make sure to send both forms from the server when upgrading the SDK for at least a full minor version.

Quinn-With-Two-Ns (approved):

PR #28, opened 2024-11-26
Hide more HTTP details in the public API

bergundy:
Before we release a stable API, we'd like to hide as much of the HTTP details from the public API.

Changes:

- Rename `Client` to `HTTPClient`
- Rename `ClientOptions` to `HTTPClientOptions`
- Rename `NewClient` to `NewHTTPClient`
- Hide `http.Response` from `UnexpectedResponseError`, it can now be obtained by casting the `Details` field
- `OperationCompletionSuccessful` now uses `nexus.Header` and `nexus.Reader` instead of `http.Header` and `io.Reader`
- `OperationCompletionUnsuccessul` now uses a `nexus.Header` instead of `http.Header`
- Rename `OperationCompletion(Un)Successful(Options)` `StartLinks` to `Links`

💥 **This is a breaking change** and requires a semver minor bump (we're still in 0 minor version) before release.

### DIFF
--- a/nexus/completion.go
+++ b/nexus/completion.go
@@ -46,8 +46,8 @@ type OperationCompletionSuccessful struct {
 	OperationID string
 	// StartTime is the time the operation started. Used when a completion callback is received before a started response.
 	StartTime time.Time
-	// StartLinks are used to link back to the operation when a completion callback is received before a started response.
-	StartLinks []Link
+	// Links are used to link back to the operation when a completion callback is received before a started response.
+	Links []Link
 }
 
 // OperationCompletionSuccessfulOptions are options for [NewOperationCompletionSuccessful].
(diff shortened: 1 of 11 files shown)


nexus/completion.go
```diff
@@ -33,10 +34,13 @@ type OperationCompletion interface {
...
-	// Body to send in the completion HTTP request.
-	// If it implements `io.Closer` it will automatically be closed by the client.
-	Body io.Reader
+	// Note that this is a Nexus header, not an HTTP header.
+	Header Header
+
+	// A [Reader] that may be directly set on the completion or constructed when instantiating via
+	// [NewOperationCompletionSuccessful].
+	// Automatically closed when the completion is delivered.
+	Reader *Reader
```
cretz, line 43: Doesn't `Reader` already have a header on it?

bergundy: Yes, that's a content header though, not a Header for the entire request.

cretz (approved):

bergundy:
> I realized I need to update the README too.

Done.

PR #29, opened 2024-12-06
Failure Converter

bergundy:
💥 BREAKING CHANGE 💥 

- Add a `FailureConverter` to convert from a `Failure` to `error` and back and remove `Failure` from the main SDK APIs.
- Changed `HandlerError` to take a `Cause error` instead of a `Failure` object.
- `HandlerError` gets an `Unwrap()` method for use with helpers in the `errors` package.
- Changed `UnsuccessfulOperationError` to take a `Cause error` instead of a `Failure` object.
- `UnsuccessfulOperationError` gets an `Unwrap()` method for use with helpers in the `errors` package.
- Added new shorthand constructors for `UnsuccessfulOperationError`: `NewFailedOperationError` and `NewCanceledOperationError`.
- Added a new `NewOperationCompletionUnsuccessful` constructor that takes an error object and options to support a failure converter.
- Rename `StartLinks` to `Links`.

### DIFF
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -65,6 +65,18 @@ type Failure struct {
 	Details json.RawMessage `json:"details,omitempty"`
 }
 
+// An error that directly represents a wire representation of [Failure].
+// The SDK will convert to this error by default unless the [FailureConverter] instance is customized.
+type FailureError struct {
+	// The underlying Failure object this error represents.
+	Failure Failure
+}
+
+// Error implements the error interface.
+func (e *FailureError) Error() string {
+	return e.Failure.Message
+}
+
 // UnsuccessfulOperationError represents "failed" and "canceled" operation results.
 type UnsuccessfulOperationError struct {
 	// State of the operation. Only [OperationStateFailed] and [OperationStateCanceled] are valid.
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -96,7 +96,7 @@ func (e *UnsuccessfulOperationError) Error() string {
 	if e.Cause == nil {
 		return fmt.Sprintf("operation %s", e.State)
 	}
-	return fmt.Sprintf("operation %s: %s", e.State, e.Error())
+	return fmt.Sprintf("operation %s: %s", e.State, e.Cause.Error())
 }
 
 // Unwrap returns the cause for use with utilities in the errors package.
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -70,8 +70,7 @@ type UnsuccessfulOperationError struct {
 	// State of the operation. Only [OperationStateFailed] and [OperationStateCanceled] are valid.
 	State OperationState
 	// The underlying cause for this error.
-	Cause      error
-	rawFailure *Failure
+	Cause error
 }
 
 // NewFailedOperationError is shorthand for constructing an [UnsuccessfulOperationError] with State set to
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -53,7 +53,9 @@ const (
 	StatusUpstreamTimeout = 520
 )
 
-// A Failure represents failed handler invocations as well as `failed` or `canceled` operation results.
+// A Failure represents failed handler invocations as well as `failed` or `canceled` operation results. Failures
+// shouldn't typically be constructed directly. The SDK APIs take a [FailureConverter] instance that can translate
+// language errors to and from [Failure] instances.
 type Failure struct {
 	// A simple text message.
 	Message string `json:"message"`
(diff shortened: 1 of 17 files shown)


nexus/api.go
```diff
@@ -65,16 +67,41 @@ type Failure struct {
...
+// NewFailedOperationError is shorthand for constructing an [UnsuccessfulOperationError] with State set to
+// [OperationStateFailed] and the given err as the Cause.
+func NewFailedOperationError(err error) *UnsuccessfulOperationError {
+	return &UnsuccessfulOperationError{
+		State: OperationStateFailed,
+		Cause: err,
+	}
+}
+
+// NewFailedOperationError is shorthand for constructing an [UnsuccessfulOperationError] with State set to
```
cretz, line 85:
```suggestion
// NewCanceledOperationError is shorthand for constructing an [UnsuccessfulOperationError] with State set to
```

nexus/client.go
```diff
@@ -27,8 +27,11 @@ type HTTPClientOptions struct {
 	// Defaults to [http.DefaultClient.Do].
 	HTTPCaller func(*http.Request) (*http.Response, error)
 	// A [Serializer] to customize client serialization behavior.
-	// By default the client handles, JSONables, byte slices, and nil.
+	// By default the client handles JSONables, byte slices, and nil.
 	Serializer Serializer
+	// A [FailureConverter] to convert a [Failure] instance to and from an [error].
+	// By default the client translates only error messages, losing type information and struct fields.
+	FailureConverter FailureConverter
```
cretz, line 34: Arguably could be called a `FailureSerializer`, but I'm ok with this too

nexus/serializer.go
```diff
@@ -226,3 +240,22 @@ type compositeSerializer struct {
...
+		Message: err.Error(),
+	}
+}
+
+// FailureToError implements FailureConverter.
+func (e errorMessageFailureConverter) FailureToError(f Failure) error {
+	return errors.New(f.Message)
+}
+
+var defaultFailureConverter FailureConverter = errorMessageFailureConverter{}
```
cretz, line 261: People should be allowed to obtain this (and the default serializer) in case they want to delegate to it from their own converters. Would recommend an exported `DefaultFailureConverter() FailureConverter` function (and `DefaultSerializer()` equivalent).

nexus/serializer.go
```diff
@@ -226,3 +240,22 @@ type compositeSerializer struct {
...
+		return Failure{}
+	}
+	return Failure{
+		Message: err.Error(),
+	}
+}
+
+// FailureToError implements FailureConverter.
+func (e errorMessageFailureConverter) FailureToError(f Failure) error {
+	return errors.New(f.Message)
```
cretz, line 258: I think we should have some kind error type here that provides the user access to the failure details. Otherwise, are we expecting Nexus clients to be forced to create their own failure converters just to see "details" on failures?

nexus/client.go
```diff
@@ -402,43 +409,49 @@ func failureFromResponse(response *http.Response, body []byte) (Failure, error)
...
 	switch response.StatusCode {
 	case http.StatusBadRequest:
-		failure := failureFromResponseOrDefault(response, body, "bad request")
-		return &HandlerError{Type: HandlerErrorTypeBadRequest, Failure: &failure}
+		failureErr := c.failureErrorFromResponseOrDefault(response, body, "bad request")
+		return &HandlerError{Type: HandlerErrorTypeBadRequest, Cause: failureErr}
 	case http.StatusUnauthorized:
-		failure := failureFromResponseOrDefault(response, body, "unauthenticated")
-		return &HandlerError{Type: HandlerErrorTypeUnauthenticated, Failure: &failure}
+		failureErr := c.failureErrorFromResponseOrDefault(response, body, "unauthenticated")
```
Quinn-With-Two-Ns, line 432: Is it intentional that `StatusUnauthorized` maps to `unauthenticated`, naively I would think the mapping would be `StatusUnauthorized` -> `unauthorized`?

bergundy: We map forbidden to unauthorized. I stole it from gRPC:

https://github.com/grpc-ecosystem/grpc-gateway/blob/a7cf811e6ffabeaddcfb4ff65602c12671ff326e/runtime/errors.go#L52C1-L55C33

Quinn-With-Two-Ns (approved):

nexus/serializer.go
cretz: Vars are mutable

nexus/serializer.go
cretz: Not following exactly. I understand complex handler side, but I am talking about client side. If I am a simple Nexus caller that wants the error details, how do I get them?

bergundy: What I ended doing in the server code base is to define a `FailureError` struct that's both an `error` and embeds a `Failure` object and a custom failure converter.

I'm just trying to discourage users from using failures directly and would rather hide them from the API until there's strong indication that they should be exposed. `Failure` is a wire level construct and shouldn't be relevant to application code, it's mostly for platforms and proxies like Temporal.

nexus/serializer.go
bergundy: We have plenty of exported vars, why are these ones more important than others to prevent mutations on?

cretz: We don't have plenty in this repo (I would have discouraged it). The one we do have is for an error which is commonly known in Go not to mutate. We should not expose the default converter as a var though since we access it in a non-thread-safe way. It's harmless to expose as a parameterless function.

nexus/serializer.go
cretz: It's really the details I'm concerned about. They should not only be for platforms/proxies IMO, same as they are not for gRPC or Temporal or whatever. No strong opinions on how you expose setting/retrieving of details from the error.

cretz (approved):

PR #32, opened 2025-01-14
Fix sync response link populating

pdoerner:


bergundy (approved):
Ah good catch! Please make sure there's test coverage.

PR #34, opened 2025-01-24
Idiomatic Operation interface

bergundy:
💥 BREAKING CHANGE 💥 

- Change the `Operation` interface and `OperationHandle` struct `GetResult` and `GetInfo` methods to `Result` and `Info` , which are more idiomatic for Go.
- Add a `MustRegister` method on `Service` for convenience.

This breaking change should have minimal impact for Temporal use cases since these two methods were never implemented in that context.

### DIFF
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -76,7 +76,7 @@ type Operation[I, O any] interface {
 	// started. Return an [OperationError] to indicate that an operation completed as failed or
 	// canceled.
 	Start(context.Context, I, StartOperationOptions) (HandlerStartOperationResult[O], error)
-	// GetResult handles requests to get the result of an asynchronous operation. Return non error result to respond
+	// Result handles requests to get the result of an asynchronous operation. Return non error result to respond
 	// successfully - inline, or error with [ErrOperationStillRunning] to indicate that an asynchronous operation is
 	// still running. Return an [OperationError] to indicate that an operation completed as failed or
 	// canceled.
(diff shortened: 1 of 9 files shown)


nexus/operation.go
```diff
@@ -88,9 +88,9 @@ type Operation[I, O any] interface {
 	// It is the implementor's responsiblity to respect the client's wait duration and return in a timely fashion.
 	// Consider using a derived context that enforces the wait timeout when implementing this method and return
 	// [ErrOperationStillRunning] when that context expires as shown in the [Handler] example.
-	GetResult(context.Context, string, GetOperationResultOptions) (O, error)
-	// GetInfo handles requests to get information about an asynchronous operation.
-	GetInfo(context.Context, string, GetOperationInfoOptions) (*OperationInfo, error)
+	Result(context.Context, string, GetOperationResultOptions) (O, error)
+	// Info handles requests to get information about an asynchronous operation.
+	Info(context.Context, string, GetOperationInfoOptions) (*OperationInfo, error)
```
cretz, line 93: I would agree this would be more idiomatic if the result and info were already present, because Go does not like `Get`-prefixed getters. But these are not getters, these are actionable items, and IMO need their verbs. If you need to call it `Fetch` or something because of an (IMO unwarranted) aversion to the word `Get`, ok.

bergundy: Hmm... I can buy that point. Maybe this is just me over-fixating on the effective Go recommendation. I sampled some client libraries and couldn't find conclusive examples of actionable APIs that follow one pattern or the other.

I buy your argument though, I would be okay leaving the API as-is.

bergundy:
Going ahead and closing this PR based on @cretz's comment.

PR #37, opened 2025-02-02
Add HandlerError.RetryBehavior

bergundy:
See also: https://github.com/nexus-rpc/api/pull/15


cretz (approved):
Is there a Java equivalent that needs to be written?

PR #41, opened 2025-02-05
Add middleware/interceptors

Quinn-With-Two-Ns:
Taken over by @bergundy (thanks @Quinn-With-Two-Ns for kicking this off).

💥 BREAKING CHANGE 💥  The experimental `WithHandlerContext` method signature was changed to accept the associated `HandlerInfo`.

Do not merge until the Temporal SDK work that depends on this is ready for review.

### DIFF
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -64,13 +64,21 @@ type RegisterableOperation interface {
 //
 // Operation implementations must embed the [UnimplementedOperation].
 //
-// All Operation methods can return a [HandlerError] to fail requests with a custom [HandlerErrorType] and structured [Failure].
-// Arbitrary errors from handler methods are turned into [HandlerErrorTypeInternal],their details are logged and hidden
-// from the caller.
+// See [OperationHandler] for more information.
 type Operation[I, O any] interface {
 	RegisterableOperation
 	OperationReference[I, O]
+	OperationHandler[I, O]
+}
 
+// OperationHandler is the interface for the core operation methods. OperationHandler implementations must embed
+// [UnimplementedOperation].
+//
+// All Operation methods can return a [HandlerError] to fail requests with a custom [HandlerErrorType] and structured [Failure].
+// Arbitrary errors from handler methods are turned into [HandlerErrorTypeInternal], when using the Nexus SDK's
+// HTTP handler, their details are logged and hidden from the caller. Other handler implementations may expose internal
+// error information to callers.
+type OperationHandler[I, O any] interface {
 	// Start handles requests for starting an operation. Return [HandlerStartOperationResultSync] to respond
 	// successfully - inline, or [HandlerStartOperationResultAsync] to indicate that an asynchronous operation was
 	// started. Return an [OperationError] to indicate that an operation completed as failed or
@@ -101,6 +109,8 @@ type Operation[I, O any] interface {
 	//  ignored by the underlying operation implemention.
 	//  2. idempotent - implementors should ignore duplicate cancelations for the same operation.
 	Cancel(ctx context.Context, token string, options CancelOperationOptions) error
+
+	mustEmbedUnimplementedOperation()
 }
 
 type syncOperation[I, O any] struct {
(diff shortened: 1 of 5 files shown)


nexus/operation.go
```diff
@@ -15,6 +15,31 @@ import (
 //	)}
 type NoValue *struct{}
 
+// OperationOptions contains the general options for an operation, across different handlers.
+//
+// NOTE: Experimental
+type OperationOptions struct {
```
bergundy, line 21: Not loving the name `Options` but I can see that it's consistent across all methods and every method has its own `Options` struct.

Maybe we should call this `MiddlewareOptions` or `MiddlewareInfo`?

nexus/operation.go
```diff
@@ -15,6 +15,31 @@ import (
...
+	// ServiceName is the name of the service that contains the operation.
+	ServiceName string
+	// OperationName is the name of the operation.
+	OperationName string
```
bergundy, line 25: I think we can drop the `Name` suffix here but don't have a strong opinion.

nexus/operation.go
```diff
@@ -70,7 +95,10 @@ type RegisterableOperation interface {
 type Operation[I, O any] interface {
 	RegisterableOperation
 	OperationReference[I, O]
+	OperationInvoker[I, O]
+}
 
+type OperationInvoker[I, O any] interface {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Oh yeah we have the name `OperationHandler ` available in Go, that is a better name

nexus/operation.go
Quinn-With-Two-Ns: Alternatively we could call this something like `CommonOperationOptions` and embedded it in every other operation struct

bergundy: No need to embed it, the options are used both in the client and the server and this information would be redundant.

bergundy: Another option is `HandlerInfo`, does that seem better to you?

Quinn-With-Two-Ns: hmm it is OK

nexus/operation.go
Quinn-With-Two-Ns: Done

nexus/operation.go
```diff
@@ -225,76 +264,126 @@ func (r *ServiceRegistry) NewHandler() (Handler, error) {
...
 
-	services map[string]*Service
+	services    map[string]*Service
+	middlewares []MiddlewareFunc
 }
 
-// CancelOperation implements Handler.
-func (r *registryHandler) CancelOperation(ctx context.Context, service, operation string, token string, options CancelOperationOptions) error {
-	s, ok := r.services[service]
+func (r *registryHandler) getOperation(options OperationOptions) (OperationInvoker[any, any], error) {
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Done

nexus/operation.go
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Done

nexus/operation.go
```diff
@@ -225,76 +265,126 @@ func (r *ServiceRegistry) NewHandler() (Handler, error) {
...
+		Operation: operation,
+		Header:    options.Header,
+	})
+	if err != nil {
+		return err
+	}
+	return h.Cancel(ctx, operationID, options)
+}
+
+// operationHandlerInfo implements Handler.
```
bergundy, line 364: nit: fix the docstring?

nexus/operation.go
```diff
@@ -225,76 +265,126 @@ func (r *ServiceRegistry) NewHandler() (Handler, error) {
...
+		Operation: operation,
+		Header:    options.Header,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return h.GetInfo(ctx, operationID, options)
+}
+
+// operationHandlerResult implements Handler.
```
bergundy, line 377: and here.

bergundy (approved):
LGTM, left a couple of minor comments.

nexus/operation.go
```diff
@@ -15,6 +15,31 @@ import (
...
+	// Header keys with the "content-" prefix are reserved for [Serializer] headers and should not be set in the
+	// client API; they are not available to server [Handler] and [Operation] implementations.
+	Header Header
+}
+
+// MiddlewareFunc is a function which receives an OperationHandler and returns another OperationHandler.
+// If the middleware wants to stop the chain before any handler is called, it can return an error.
+//
+// NOTE: Experimental
+type MiddlewareFunc func(HandlerInfo, OperationHandler[any, any]) (OperationHandler[any, any], error)
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: Hmm, so the `HandlerInfo ` contains the headers that the respective handler inputs already have. Would we be OK moving them out of the handler input into the `HandlerInfo `?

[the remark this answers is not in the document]
bergundy: I'm on the fence here. I kept going back and forth between headers in the context and explicit argument and that's where I landed. We should discuss this offline before we stabilize the API.

Quinn-With-Two-Ns (commented):
I can' approve because this is my own PR but LGTM

bergundy:
Verified with the Temporal SDK, good to merge now.

PR #42, opened 2025-02-05
Prep for stable API

bergundy:
- [Remove EXPERIMENTAL warning from README and mark specific exports as experimental](https://github.com/nexus-rpc/sdk-go/commit/694a754acc5cd250f7c7d31026f510d0d7e5783f)
- [Sanitize token header](https://github.com/nexus-rpc/sdk-go/commit/1f6799bd3fa6188b754ad75bc42c3c0d56adaa17)
- [Add AddHandlerLinks method and deprecate links on result objects](https://github.com/nexus-rpc/sdk-go/commit/8f448a7ae4054a27b38b1181bd264aa0096a9e44)

### DIFF
--- a/nexus/server.go
+++ b/nexus/server.go
@@ -26,7 +26,7 @@ type handlerCtx struct {
 	links []Link
 }
 
-// AddHandlerLinks associates a link to with the current operation, which is propagated back to the caller. This method
+// AddHandlerLinks associates links with the current operation to be propagated back to the caller. This method
 // Can be called from an [Operation] handler Start method or from a [Handler] StartOperation method. The context
 // provided must be the context passed to the handler. This method may be called multiple times for a given handler,
 // each call appending additional links. Links will only be attached on successful responses.
--- a/nexus/server.go
+++ b/nexus/server.go
@@ -449,9 +449,19 @@ func (h *httpHandler) handleRequest(writer http.ResponseWriter, request *http.Re
 		h.writeFailure(writer, HandlerErrorf(HandlerErrorTypeBadRequest, "failed to parse URL path"))
 		return
 	}
+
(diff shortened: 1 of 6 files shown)


nexus/server.go
```diff
@@ -13,9 +13,30 @@ import (
...
+type handlerCtxKeyType struct{}
+
+var handlerCtxKey = handlerCtxKeyType{}
+
+type handlerCtx struct {
+	mu    sync.Mutex
+	links []Link
+}
+
+// AddHandlerLinks associates a link to with the current operation, which is propagated back to the caller. This method
```
bergundy, line 29:
```suggestion
// AddHandlerLinks associates links with the current operation to be propagated back to the caller. This method
```

Quinn-With-Two-Ns (approved):

nexus/server.go
```diff
@@ -13,9 +13,30 @@ import (
...
+type handlerCtx struct {
+	mu    sync.Mutex
+	links []Link
+}
+
+// AddHandlerLinks associates links with the current operation to be propagated back to the caller. This method
+// Can be called from an [Operation] handler Start method or from a [Handler] StartOperation method. The context
+// provided must be the context passed to the handler. This method may be called multiple times for a given handler,
+// each call appending additional links. Links will only be attached on successful responses.
+func AddHandlerLinks(ctx context.Context, links ...Link) {
```
cretz, line 33: I should be able to see links that exist, remove links, etc. I would recommend at least a getter and setter if not access to a structure that contains links as a field.

bergundy: Sorry, I already merged, I can address in a followup PR if needed.

I think it's okay to expose a simple API for now, following https://pkg.go.dev/google.golang.org/grpc#SetHeader, and add more functionality later.

PR #44, opened 2025-02-06
Add methods to create a handler context and extract links from context

bergundy:
Follow up to #42, this is so we can use `AddHandlerLinks` in the Temporal SDK.

### DIFF
--- a/nexus/server.go
+++ b/nexus/server.go
@@ -26,6 +26,21 @@ type handlerCtx struct {
 	links []Link
 }
 
+// WithHandlerContext returns a new context from a given context setting it up for being used for handler methods.
+// Meant to be used by frameworks, not directly by applications.
+func WithHandlerContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, handlerCtxKey, &handlerCtx{})
+}
+
+// HandlerLinks retrieves the attached links on the given handler context. The returned slice should not be mutated.
+// The context provided must be the context passed to the handler.
+func HandlerLinks(ctx context.Context) []Link {
+	hctx := ctx.Value(handlerCtxKey).(*handlerCtx)
+	hctx.mu.Lock()
+	defer hctx.mu.Unlock()
+	return hctx.links
+}
+
 // AddHandlerLinks associates links with the current operation to be propagated back to the caller. This method
 // Can be called from an [Operation] handler Start method or from a [Handler] StartOperation method. The context
 // provided must be the context passed to the handler. This method may be called multiple times for a given handler,
--- a/nexus/server.go
+++ b/nexus/server.go
@@ -34,30 +34,33 @@ func WithHandlerContext(ctx context.Context) context.Context {
 	return context.WithValue(ctx, handlerCtxKey, &handlerCtx{})
 }
 
-// InHandlerContext returns true if the given context is a handler context where [AddHandlerLinks] and [HandlerLinks]
+// IsHandlerContext returns true if the given context is a handler context where [AddHandlerLinks] and [HandlerLinks]
 // can be called. It will only return true when called from an [Operation] handler Start method or from a [Handler]
 // StartOperation method.
 //
 // NOTE: Experimental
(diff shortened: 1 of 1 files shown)


nexus/server.go
```diff
@@ -26,6 +26,21 @@ type handlerCtx struct {
...
+	return context.WithValue(ctx, handlerCtxKey, &handlerCtx{})
+}
+
+// HandlerLinks retrieves the attached links on the given handler context. The returned slice should not be mutated.
+// The context provided must be the context passed to the handler.
+func HandlerLinks(ctx context.Context) []Link {
+	hctx := ctx.Value(handlerCtxKey).(*handlerCtx)
+	hctx.mu.Lock()
+	defer hctx.mu.Unlock()
+	return hctx.links
```
cretz, line 41: I think these should be copied on return (new slice/array + https://pkg.go.dev/builtin#copy is quite cheap especially for small uses) and there should be a set links too

nexus/server.go
```diff
@@ -26,6 +26,21 @@ type handlerCtx struct {
...
+// WithHandlerContext returns a new context from a given context setting it up for being used for handler methods.
+// Meant to be used by frameworks, not directly by applications.
+func WithHandlerContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, handlerCtxKey, &handlerCtx{})
+}
+
+// HandlerLinks retrieves the attached links on the given handler context. The returned slice should not be mutated.
+// The context provided must be the context passed to the handler.
+func HandlerLinks(ctx context.Context) []Link {
+	hctx := ctx.Value(handlerCtxKey).(*handlerCtx)
```
cretz, line 38:
```suggestion
	hctx, ok := ctx.Value(handlerCtxKey).(*handlerCtx)
	if !ok {
		return nil
	}
```

Or if panic is intentional, document that it can panic if called outside of context (and possibly provide a way to know whether you're in a handler context). Same comment applies to `AddHandlerLinks`.

bergundy: I can document, the panic is intentional.

cretz: We should consider a utility to let someone know a context is a handler context then. Basically general utility users should be able to know whether something they are about to call will panic (and therefore can prevent it with a conditional)

nexus/server.go
bergundy: Why? Are you worried about users mutating? Is it typical for Go libraries to protect users this way?

nexus/server.go
bergundy: Sure, that makes sense.

pdoerner (approved):

nexus/server.go
cretz: Yes and yes, e.g. https://github.com/grpc/grpc-go/blob/v1.70.0/metadata/metadata.go#L197. This is a very common approach for returning mutable slices that are rarely asked for.

nexus/server.go
```diff
@@ -26,17 +26,66 @@ type handlerCtx struct {
...
+// used to verify the context is valid.
+//
+// NOTE: Experimental
+func HandlerLinks(ctx context.Context) []Link {
+	hctx := ctx.Value(handlerCtxKey).(*handlerCtx)
+	hctx.mu.Lock()
+	links := hctx.links
+	hctx.mu.Unlock()
+	cpy := make([]Link, len(links))
+	copy(cpy, links)
```
cretz, line 57:
```suggestion
	cpy := make([]Link, len(links))
	copy(cpy, links)
	hctx.mu.Unlock()
```

The lock needs to be held while it is accessed

bergundy: Why? We always assign to `h.links`, we never mutate it directly.

cretz: It's a common misunderstanding in Go that `append` doesn't mutate the first param

bergundy: Thanks, I wasn't aware.

cretz (approved):

PR #46, opened 2025-02-21
Update version to 0.3.0

bergundy:


Quinn-With-Two-Ns:
Will this release be OK with the current Go SDK release

bergundy:
Yes, I posted this internally:

> Don't mind the breaking change, this API was only used in Temporal Go SDK master, not in a released version.

Quinn-With-Two-Ns (approved):

PR #48, opened 2025-04-21
Add contrib/nexusproto package for protobuf JSON and binary serialization support

bergundy:
Note that the separate package was made to avoid adding `google.golang.org/protobuf` as a dependency for every SDK user.


pdoerner (approved):

contrib/nexusproto/serializer.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return nil, err
+	}
+	messageType := string(msg.ProtoReflect().Descriptor().FullName())
+	return &nexus.Content{
+		Header: nexus.Header{
+			"type": mime.FormatMediaType("application/json", map[string]string{
+				"format":       "protobuf",
+				"message-type": messageType,
+			}),
```
cretz, line 75: Can we update the Nexus spec with these details for supporting protobuf?

contrib/nexusproto/serializer.go
```diff
@@ -0,0 +1,159 @@
...
+		},
+		Data: data,
+	}, nil
+}
+
+type protoBinarySerializer struct{}
+
+var ErrInvalidProtoObject = errors.New("invalid protobuf object")
```
cretz, line 83: I don't think this needs to be part of the public library API

contrib/nexusproto/serializer.go
```diff
@@ -0,0 +1,159 @@
...
+	// In the second case, we need to instantiate an empty proto struct:
+	if elem.Kind() == reflect.Ptr {
+		empty := reflect.New(elem.Elem())
+		rv.Elem().Set(empty)
+		rv = empty
+	}
+
+	msg, ok := rv.Interface().(proto.Message)
+	if !ok {
+		return nil, nexus.ErrSerializerIncompatible
```
cretz, line 33: I see we are using the same error for serialize and deserialize, even though the error for serialize is meant to be a benign error and deserialize is meant to be a user-facing error, is that correct? I think we should use a more descriptive error here saying the content is a protobuf message but the type is not.

contrib/nexusproto/serializer.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (s protoJsonSerializer) Deserialize(c *nexus.Content, v any) error {
+	if !s.isValidContentType(c.Header["type"]) {
+		return nexus.ErrSerializerIncompatible
+	}
+	msg, err := messageFromAny(v)
+	if err != nil {
+		return err
+	}
+	return protojson.Unmarshal(c.Data, msg)
```
cretz, line 57: Before this, I think we should validate that the `message-type` is the same as the message type

contrib/nexusproto/serializer.go
cretz, line 1: Are there issues open for the other Nexus SDKs (or in general just some way not to lose track of these new things that are only being done in one language)?

contrib/nexusproto/serializer.go
bergundy: Sure, I can do that.

contrib/nexusproto/serializer.go
bergundy: Agree, let me remove it.

contrib/nexusproto/serializer.go
bergundy: For serialize it may not be benign if we exhaust all serializers. I will add more info to the message.

contrib/nexusproto/serializer.go
bergundy: On it.

contrib/nexusproto/serializer.go
bergundy: I can open issues.

contrib/nexusproto/serializer.go
bergundy: https://github.com/nexus-rpc/api/pull/18

contrib/nexusproto/serializer.go
bergundy: https://github.com/nexus-rpc/sdk-java/issues/27
https://github.com/nexus-rpc/sdk-python/issues/2
https://github.com/nexus-rpc/sdk-typescript/issues/2

contrib/nexusproto/serializer.go
```diff
@@ -0,0 +1,174 @@
...
+		return ""
+	}
+	return params["message-type"]
+}
+
+func (s protoJsonSerializer) Deserialize(c *nexus.Content, v any) error {
+	messageTypeFromHeader := s.extractMessageType(c.Header["type"])
+	if messageTypeFromHeader == "" {
+		return fmt.Errorf("%w: incompatible content type header", nexus.ErrSerializerIncompatible)
+	}
```
cretz, line 55: I thought the incompatible-serializer made a bit of sense here. Can you help me understand fall-through behavior on deserialize? It seems we don't pick the deserializer _explicitly_ based on the mime type, so we cycle through right? So if I have both json and binary enabled, but json preferred, does this error cause it to fall through to binary? Ideally we only want some errors to fall through (e.g. incompatible), but other errors to be bubbled immediately without trying other deserializers.

bergundy: Fallthrough is only on ErrSerializerIncompatible.

bergundy: Yup, that's what is happening here.

cretz: Ah, I see, "Fallthrough is only on ErrSerializerIncompatible and any error with that in the chain"

cretz (approved):
LGTM, but I would like @Quinn-With-Two-Ns to look too since this is our first `contrib` library of Nexus and I want to confirm that we are ok with this approach.

contrib/nexusproto/serializer.go
```diff
@@ -0,0 +1,174 @@
...
+	msg, err := messageFromAny(v)
+	if err != nil {
+		return err
+	}
+	messageTypeFromValue := string(msg.ProtoReflect().Type().Descriptor().FullName())
+	if messageTypeFromHeader != messageTypeFromValue {
+		return fmt.Errorf("serialized message type: %q is different from the provided value: %q", messageTypeFromHeader, messageTypeFromValue)
```
Quinn-With-Two-Ns, line 113: This is different from how the Go SDK handles proto deserialization, the Go SDK doesn't require the type match

Quinn-With-Two-Ns (approved):
No concern with this approach to optional dependencies, we don't have to call it `contrib` but I am fine with that name.

The converter does behave a bit differently then the Temporal Go SDK protobuf converter, not clear if that is intentional or we want them to be consistent?

contrib/nexusproto/serializer.go
bergundy: I think that may be since the Go SDK added message type later on and the message type header isn't considered required.

cretz:
I like calling it `contrib` so we don't have modules sprinkled throughout, only under `contrib`. I like the name better than "extensions" or "plugins" or "integrations" or whatever. Some libs use completely separate repos (e.g. https://github.com/open-telemetry/opentelemetry-go-contrib) but I don't think we'll have enough to justify it.

PR #49, opened 2025-05-09
Bump version to 0.4.0

bergundy:
This version only includes a couple of doc fixes and the addition of the contrib proto package. I think it deserves a minor due to this additional package added.

### DIFF
--- a/nexus/api.go
+++ b/nexus/api.go
@@ -18,7 +18,7 @@ import (
 )
 
 // Package version.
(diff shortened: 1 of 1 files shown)


bergundy:
Scratch that, there are newly exported variables in 0.4.0 that are needed in the proto package. Reopened.

rodrigozhou (approved):

PR #52, opened 2025-07-09
Fix client content length header behavior

bergundy:
Before this change a request would use `Transfer-Encoding: chunked` and the `Content-Length` header would not be added.


bergundy:
I will leave this open and let @pdoerner merge it post client refactor.

PR #53, opened 2025-07-11
Refactor Nexus client interface

pdoerner:
Refactored the Nexus client interface. We now have a top level `Client` which accepts the new `Transport` interface for making network calls. The previous `HTTPClient` implementation has been refactored into a `Transport` implementation. Similarly, the top level `OperationHandle` struct now accepts a `Transport` which it will use to make network calls.

There were several changes to the signatures exposed by both `Client` and `OperationHandle`, most notably `OperationHandle` now has methods to return full response objects and the old methods are now sugar on top of these.


.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
pdoerner, line 1: This was just annoying me but I can leave it off the PR if it is not wanted.

nexus/transport.go
```diff
@@ -0,0 +1,707 @@
...
+
+		// GetOperationInfo returns information about a specific operation. Should only be called through an
+		// [OperationHandle].
+		//
+		// NOTE: Experimental
+		GetOperationInfo(
+			ctx context.Context,
+			service string,
+			operation string,
+			token string,
```
cretz, line 66: I think we may hit a situation where transport call options are not necessarily the same as their high-level counterparts. These 3 are transport-only options. But they are not in a structure where we can ever add more transport-only options, so they are the last 3 ever. I think we should consider structures specific to the transport calls (can prefix their to delineate) even at the acceptable cost of more types/conversion. The transport options can carry the high-level options struct as a field, don't have to expand out field for field.

nexus/transport.go
```diff
@@ -0,0 +1,707 @@
...
+			options CancelOperationOptions,
+		) (*CancelOperationResponse, error)
+	}
+
+	// StartOperationResponse is the response to Transport.StartOperation calls. One and only one of Complete or
+	// Pending will be populated.
+	//
+	// NOTE: Experimental
+	StartOperationResponse[T any] struct {
```
cretz, line 118: IMO all of this stuff could be prefixed with `Transport` and make it transport only. A clean separation between the transport impl and the actual client impl has value IMO compared to reusing transport types in non-transport situations or cluttering the API with transport-only types unqualified.

cretz (commented):
Overall LGTM, only minor things

bergundy:
Some other things to consider for before we finalize this work (not necessarily in this PR):

1. Do we want to modify the timeout semantics? E.g. can we just use context deadline?
2. Do we want to split out into separate packages?
3. Are we okay keeping the same options structs for handlers and clients?
4. Where does serialization happen? In the transport or closer to the application (e.g. the client and handler level vs. the server and transport level).

nexus/transport.go
```diff
@@ -0,0 +1,707 @@
...
+
+const getResultContextPadding = time.Second * 5
+
+var errOperationWaitTimeout = errors.New("operation wait timeout")
+
+type (
+	// Transport is the low-level abstraction used by [Client] and [OperationHandle] to make network calls.
+	//
+	// NOTE: Experimental
+	Transport interface {
```
cretz, line 28: I think this needs a `Close()`. This brings up a philosophical discussion where we have to ask whether `Client` needs a `Close` and does it call `Transport.Close`. My opinion is, if you are going to keep the create-client-and-transport helper, you need a `Client.Close` that only invokes `Transport.Close` if it was created with the client by our code, otherwise it's a no-op (could do counters, but that's too much magic). If you are not going to keep the create-client-and-transport helper and instead require two-step instantiation, you don't need a `Client.Close` at all. I lean the latter since the Client returned from Temporal SDK doesn't need to be `Close`d (because it's outer client is responsible for that).

nexus/handle.go
```diff
@@ -21,52 +17,51 @@ type OperationHandle[T any] struct {
...
-	if err != nil {
-		return nil, err
-	}
+// GetInfoResponse gets operation information, issuing a network request to the service handler. It returns a
+// GetOperationInfoResponse object containing operation information and metadata.
+//
+// NOTE: Experimental
+func (h *OperationHandle[T]) GetInfoResponse(ctx context.Context, options GetOperationInfoOptions) (*GetOperationInfoResponse, error) {
+	return h.transport.GetOperationInfo(ctx, h.Service, h.Operation, h.Token, options)
+}
```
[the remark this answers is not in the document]
pdoerner: Yes, we expect callers will want Links at the very least. Personally I think the Go "overload" is fine and was agreed on last time we discussed offline. I have adjusted the names and return types, if that helps you feel better about it.

nexus/handle.go
```diff
@@ -78,147 +71,79 @@ func (h *OperationHandle[T]) GetInfo(ctx context.Context, options GetOperationIn
...
-			return any(s).(T), nil
-		} else {
-			return result, s.Consume(&result)
-		}
-	}
-}
-
-func (h *OperationHandle[T]) sendGetOperationResultRequest(request *http.Request) (*http.Response, error) {
-	response, err := h.client.options.HTTPCaller(request)
+func (h *OperationHandle[T]) GetResultWithDetails(ctx context.Context, options GetOperationResultOptions) (*OperationHandleResultWithDetails[T], error) {
```
pdoerner, line 82: Considering leaving this off but will leave it in the PR as an example of how we would expose additional links + metadata on `OperationHandle` methods. If we deem it unnecessary, I will remove before merging.

nexus/transport.go
```diff
@@ -0,0 +1,675 @@
...
+
+	// StartOperation calls the configured Nexus endpoint to start an operation.
+	//
+	// This method has the following possible outcomes:
+	//
+	//  1. The operation completes successfully. The result of this call will be set as a [LazyValue] in
+	//     TransportStartOperationResponse.Complete and must be consumed to free up the underlying connection.
+	//
+	//	2. The operation completes unsuccessfully. The response will contain an [OperationError] in
+	//	   TransportStartOperationResponse.Complete
```
bergundy, line 33:
```suggestion
	// 2. The operation completes unsuccessfully. The response will contain an [OperationError] in
	//    TransportStartOperationResponse.Complete
```

nexus/transport.go
```diff
@@ -0,0 +1,675 @@
...
+	//     such as getting its result.
+	//
+	//  4. There was an error making the call. The returned response will be nil and the error will be non-nil.
+	//     Most often it will be a [HandlerError].
+	//
+	// NOTE: Experimental
+	StartOperation(ctx context.Context, input any, options TransportStartOperationOptions) (*TransportStartOperationResponse[*LazyValue], error)
+
+	// GetOperationInfo returns information about a specific operation. Should only be called through an
+	// [OperationHandle].
```
bergundy, line 46: I would consider removing these comment on what should only be called through an `OperationHandle`. This entire interface is considered low level AFAIC and should typically not be called directly.

nexus/transport.go
```diff
@@ -0,0 +1,675 @@
...
+type TransportCancelOperationResponse struct{}
+
+// OperationResult contains the final value or error returned by an operation handler. One and only one of
+// result or err will be populated. Use OperationResult.Get to retrieve the result.
+//
+// In most cases err will be an [OperationError]. Other failures, such as [HandlerError], will be returned by
+// the Transport method called to indicate a failure to communicate with the operation handler.
+//
+// NOTE: Experimental
+type OperationResult[T any] struct {
```
bergundy, line 140: Does this need to be exported?

nexus/transport.go
```diff
@@ -0,0 +1,675 @@
...
+}
+
+// User-Agent header set on HTTP requests.
+const userAgent = "Nexus-go-sdk/" + version
+const headerUserAgent = "User-Agent"
+const getResultContextPadding = time.Second * 5
+
+var errOperationWaitTimeout = errors.New("operation wait timeout")
+
+type (
```
bergundy, line 159: nit: remove the grouping.

nexus/transport.go
```diff
@@ -0,0 +1,675 @@
...
+		//
+		// NOTE: Experimental
+		UseOperationID bool
+	}
+)
+
+// NewHTTPTransport creates a new [Transport] backed by HTTP from the provided [HTTPTransportOptions].
+//
+// NOTE: Experimental
+func NewHTTPTransport(options HTTPTransportOptions) (*HTTPTransport, error) {
```
bergundy, line 195: Consider also adding a `NewHTTPClient` helper with `HTTPClientOptions` that flatten out all of the shared client and transport options.

nexus/client.go
```diff
@@ -1,349 +1,113 @@
...
-
-	return &UnexpectedResponseError{
-		Message: message,
-		Details: response,
-		Failure: failure,
-	}
-}
-
-// An HTTPClient makes Nexus service requests as defined in the [Nexus HTTP API].
+// A Client makes Nexus service requests as defined in the [Nexus API].
```
bergundy, line 16: I would remove the link to the API spec here and only put it on the HTTP transport.

bergundy (commented):
Please update the README before merging into main, you can wait on package separation if you want though.

nexus/transport.go
bergundy: Okay, we can bring it back later.

nexus/transport.go
bergundy: I think it's fine since it's it'll move into a separate `transport` package. Just might be a bit awkward to embed a transport type directly.

I would consider exposing the error and the result separately on the transport and exporting this in the main package. Not blocking the PR.

bergundy (approved):

nexus/handle.go
cretz: @bergundy - did we want to remove this before merging and revisit advanced result-with-details later?

cretz (commented):
Nothing blocking for me, just wanting to get final confirmation whether we're punting result-with-details in a high-level form or not

nexus/handle.go
pdoerner: Not planning to merge until Temporal SDK and server are ready, so we can decide at that point if we want this initially or plan to add it when asked.

bergundy: I will defer to @pdoerner to make this call.

PR #55, opened 2025-07-31
Rename service client and add UnimplementedTransport

pdoerner:
* Renamed `client.go` -> `service_client.go` and `Client` -> `ServiceClient`
* Renamed `UnexpectedResponseError` -> `TransportError` and moved to `transport.go`
* Added `UnimplementedTransport` and `mustEmbedUnimplementedTransport` to the `Transport` interface for future compatibility

### DIFF
--- a/nexus/transport.go
+++ b/nexus/transport.go
@@ -17,7 +17,9 @@ import (
 	"github.com/google/uuid"
 )
 
-// Transport is the low-level abstraction used by [Client] and [OperationHandle] to make network calls.
+// Transport is the low-level abstraction used by [ServiceClient] and [OperationHandle] to make network calls.
+//
+// Implementations must embed [UnimplementedTransport] for future compatibility.
 //
 // NOTE: Experimental
 type Transport interface {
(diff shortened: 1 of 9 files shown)


nexus/transport.go
```diff
@@ -78,6 +80,43 @@ type Transport interface {
...
+// TransportError indicates a client encountered something unexpected in the server's response.
+type TransportError struct {
+	// Error message.
+	Message string
+	// Optional failure that may have been embedded in the response.
+	Failure *Failure
+	// Additional transport specific details.
+	// For HTTP, this would include the HTTP response. The response body will have already been read into memory and
+	// does not need to be closed.
+	Details any
```
bergundy, line 99: Should we change this to `Cause error`? That would allow unwrapping to get the underlying error.

We could make a `struct HTTPResponseError { *http.Response }` helper for that.

nexus/transport.go
```diff
@@ -144,6 +183,33 @@ func (r *OperationResult[T]) Get() (T, error) {
...
 
+// UnimplementedTransport must be embedded into any [Transport] implementation for future compatibility.
+// It implements all methods on the [Transport] interface, returning unimplemented errors if they are not implemented by
+// the embedding type.
+type UnimplementedTransport struct{}
+
+func (u UnimplementedTransport) mustEmbedUnimplementedTransport() {}
+
+func (u UnimplementedTransport) StartOperation(_ context.Context, _ any, _ TransportStartOperationOptions) (*TransportStartOperationResponse[*LazyValue], error) {
+	return nil, HandlerErrorf(HandlerErrorTypeNotImplemented, "not implemented")
```
bergundy, line 194: This should be a transport error IMHO:

```suggestion
	return nil, &TransportError{
	  Message: "not implemented",
	}
```

nexus/transport.go
```diff
@@ -78,6 +80,43 @@ type Transport interface {
...
+	// does not need to be closed.
+	Details any
+}
+
+// Error implements the error interface.
+func (e *TransportError) Error() string {
+	return e.Message
+}
+
+func newTransportError(message string, response *http.Response, body []byte) error {
```
Quinn-With-Two-Ns, line 107:
```suggestion
func newHTTPTransportError(message string, response *http.Response, body []byte) error {
```

since this is for the HTTP transport right? the gRPC transport won't use this I think

nexus/transport.go
pdoerner: Whoops. Addressed now.

PR #57, opened 2025-08-07
Add StartOperation response variants

pdoerner:
Changed the format of `TransportStartOperationResponse` and `ClientStartOperationResponse` to contain a single `Variant` instead of multiple nil-able fields.

This allows for more flexibility when adding additional response types in the future and makes it easier for callers to use type assertions to determine the outcome of their StartOperation call.

For users that do not want to expand the types, `Sync()` and `Async()` (the only two current response variants) have been added so users can directly get the value, or nil if that response variant was not set.

The approach is essentially lifted from the way protobuf handles oneof variants.

### DIFF
--- a/nexus/cancel_test.go
+++ b/nexus/cancel_test.go
@@ -44,7 +44,7 @@ func TestCancel_HandleFromStart(t *testing.T) {
 
 	result, err := client.StartOperation(ctx, "f/o/o", nil, StartOperationOptions{})
 	require.NoError(t, err)
-	handle := result.Pending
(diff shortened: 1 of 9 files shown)


bergundy (changes requested):
As discussed on slack, we will not be doing this and instead will just rename `Pending` and `Successful` to `Sync` and `Async`. No need to make those methods instead of exported struct fields.

pdoerner:
Synced offline and decided this will not give us much type safety. Will stick with the multiple optional fields for now.

PR #59, opened 2025-09-18
Add CloseTime to OperationCompletion* structs

lina-temporal:
What?
--
- Added `CloseTime` to the various OperationCompletion* structs.

Why?
--
- `CloseTime` is useful in determining an order of events from completion callbacks. We already have `StartTime`, but adding `CloseTime` adds flexibility and is generally useful as for metrics.
- In CHASM Scheduler, `CloseTime` is used to determine the earliest possible time our system could kick off a buffered start in sequence of a prior start completing (drives a customer-facing metric).
- If we weren't to add this field in Nexus, we'd have to make an additional follow-up `Describe` call to determine it, as part of processing the Nexus callback.

### DIFF
--- a/nexus/completion.go
+++ b/nexus/completion.go
@@ -10,9 +10,6 @@ import (
 	"net/http"
 	"strconv"
 	"time"
-
-	"google.golang.org/protobuf/encoding/protojson"
-	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 // NewCompletionHTTPRequest creates an HTTP request that delivers an operation completion to a given URL.
@@ -156,11 +153,7 @@ func (c *OperationCompletionSuccessful) applyToHTTPRequest(request *http.Request
 		request.Header.Set(headerOperationStartTime, c.StartTime.Format(http.TimeFormat))
 	}
 	if c.Header.Get(headerOperationCloseTime) == "" && !c.CloseTime.IsZero() {
-		timestamp, err := getCloseTimeHeader(c.CloseTime)
-		if err != nil {
-			return err
-		}
-		request.Header.Set(headerOperationCloseTime, timestamp)
+		request.Header.Set(headerOperationCloseTime, c.CloseTime.Format(time.RFC3339Nano))
 	}
 	if c.Header.Get(headerLink) == "" {
 		if err := addLinksToHTTPHeader(c.Links, request.Header); err != nil {
@@ -266,11 +259,7 @@ func (c *OperationCompletionUnsuccessful) applyToHTTPRequest(request *http.Reque
 		request.Header.Set(headerOperationStartTime, c.StartTime.Format(http.TimeFormat))
 	}
 	if c.Header.Get(headerOperationCloseTime) == "" && !c.CloseTime.IsZero() {
-		timestamp, err := getCloseTimeHeader(c.CloseTime)
-		if err != nil {
-			return err
-		}
-		request.Header.Set(headerOperationCloseTime, timestamp)
+		request.Header.Set(headerOperationCloseTime, c.CloseTime.Format(time.RFC3339Nano))
 	}
 	if c.Header.Get(headerLink) == "" {
 		if err := addLinksToHTTPHeader(c.Links, request.Header); err != nil {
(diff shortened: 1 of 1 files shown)


nexus/completion.go
```diff
@@ -176,6 +184,8 @@ type OperationCompletionUnsuccessful struct {
 	OperationToken string
 	// StartTime is the time the operation started. Used when a completion callback is received before a started response.
 	StartTime time.Time
+	// CloseTime is the time the operation completed. Used when a completion callback is received before a started response.
```
bergundy, line 187:
```suggestion
	// CloseTime is the time the operation completed. This may be different from the time the completion callback is delivered.
```

nexus/completion.go
```diff
@@ -198,6 +208,8 @@ type OperationCompletionUnsuccessfulOptions struct {
...
```
bergundy, line 211:
```suggestion
	// CloseTime is the time the operation completed. This may be different from the time the completion callback is delivered.
```

nexus/completion_test.go
bergundy, line 1: Could you please provide the expected start time and complete time to `successfulCompletionHandler` and verify that the times match the `CompletionRequest` fields?

nexus/completion.go
```diff
@@ -277,6 +293,8 @@ type CompletionRequest struct {
...
```
bergundy, line 296:
```suggestion
	// CloseTime is the time the operation completed. This may be different from the time the completion callback is delivered.
```

nexus/completion_test.go
lina-temporal: Added, I was a bit surprised that we only get back seconds resolution for this due to our use of RFC 882/1123 for header time formats, but we can chat about that elsewhere.

nexus/completion.go
```diff
@@ -147,6 +152,9 @@ func (c *OperationCompletionSuccessful) applyToHTTPRequest(request *http.Request
 	if c.Header.Get(headerOperationStartTime) == "" && !c.StartTime.IsZero() {
 		request.Header.Set(headerOperationStartTime, c.StartTime.Format(http.TimeFormat))
 	}
+	if c.Header.Get(headerOperationCloseTime) == "" && !c.CloseTime.IsZero() {
+		request.Header.Set(headerOperationCloseTime, c.CloseTime.Format(time.RFC3339Nano))
```
bergundy, line 156: @cretz we wanted at least millisecond resolution here. Do you have a different time format in mind that would be easy to use across languages?
There isn't a standard time format with this resolution in the HTTP spec AFAICT.

bergundy: We plan to deprecate start time, which uses `http.TimeFormat`, and figure out a better approach to address the race for complete-before-start so I'm not worried about the inconsistency here.

bergundy (approved):

bergundy:
> This use of `time.RFC3339Nano` should conform well to spec. Will the other SDK PRs be coming? We need to take care not to prioritize Go only.

Other SDKs haven't implemented operation completion clients yet.

nexus/completion.go
bergundy: @cretz is this comment blocking the PR?
https://github.com/nexus-rpc/api/pull/24#pullrequestreview-3291042297

bergundy:
> We should not implement it in this SDK until/unless there are plans to do it in all SDKs. If there are plans, they should be tracked somewhere (ideally they should be done at the same time or near the same time to not have prolonged SDK disparity). Surely there's no urgency to merge/release this feature/update as single-SDK only? Or IIRC to get to SDK parity, some of this client stuff is going to be deleted right? If we already have SDK disparity with this client feature, we should not be growing the disparity if we can help it, we should be reducing it.

We will move this code into the server codebase soon. We discussed not exposing servers and clients at all even in Go since we acknowledge that this API is meant for integrations and our story is not very clear here yet.

nexus/completion.go
```diff
@@ -10,6 +10,9 @@ import (
 	"net/http"
 	"strconv"
 	"time"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/timestamppb"
```
bergundy, line 15: Nexus doesn't depend on protobuf and shouldn't start depending on it now. You should copy over the code instead.

nexus/completion.go
```diff
@@ -245,6 +258,9 @@ func (c *OperationCompletionUnsuccessful) applyToHTTPRequest(request *http.Reque
 	if c.Header.Get(headerOperationStartTime) == "" && !c.StartTime.IsZero() {
 		request.Header.Set(headerOperationStartTime, c.StartTime.Format(http.TimeFormat))
 	}
+	if c.Header.Get(headerOperationCloseTime) == "" && !c.CloseTime.IsZero() {
+		request.Header.Set(headerOperationCloseTime, c.CloseTime.Format(time.RFC3339Nano))
```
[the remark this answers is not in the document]
cretz: Note, I was saying just take the few lines of code they have to make it conform. Regardless, from what I hear/hope, all of this code will be removed from the Nexus SDK soon.

PR #61, opened 2025-10-10
Add Nexus Callback Token to StartOperationOptions

chaptersix:
In support of [this feature](https://github.com/temporalio/temporal/issues/8298) this PR added the CallbackToken to the StartOperationOptions.

### DIFF
--- a/nexus/operation.go
+++ b/nexus/operation.go
@@ -85,10 +85,10 @@ type OperationHandler[I, O any] interface {
 	// canceled.
 	Start(ctx context.Context, input I, options StartOperationOptions) (HandlerStartOperationResult[O], error)
 	// Cancel handles requests to cancel an asynchronous operation.
-	// Cancellation in Nexus is:
+	// Cancelation in Nexus is:
 	//  1. asynchronous - returning from this method only ensures that cancelation is delivered, it may later be
 	//  ignored by the underlying operation implementation.
-	//  2. idempotent - implementors should ignore duplicate cancelation for the same operation.
+	//  2. idempotent - implementors should ignore duplicate cancelations for the same operation.
 	Cancel(ctx context.Context, token string, options CancelOperationOptions) error
 
 	mustEmbedUnimplementedOperation()


bergundy:
> > ```
> > In the Nexus SPEC.md, call out that the Nexus-Callback-Token is a standard header and that it is required as part of the StartOperation request.
> > ```
> > @bergundy should the sdk require the user pass in the callbacktoken?
> 
> I don't think this is relevant now. Could use some help with a better description.

It's still relevant even though we inlined the implementation in the Temporal codebase.

nexus/options.go
```diff
@@ -18,6 +18,11 @@ type StartOperationOptions struct {
 	//
 	// Implement a [CompletionHandler] and expose it as an HTTP handler to handle async completions.
 	CallbackURL string
+	// CallbackToken is a required caller-generated token that uniquely identifies the originating operation.
+	// This token is used for request routing and correlation between callback requests and their originating operations.
+	// Handlers must Include the CallbackToken in all requests to the caller's CallbackURL via the "token" header.
```
bergundy:
```suggestion
	// CallbackToken is a caller-generated token that is used to recreate context when processing the callback.
	// CallbackToken must be provided if CallbackURL is not empty.
	// Handlers must Include the CallbackToken when they invoke the callback.
```

nexus/operation.go
```diff
@@ -85,10 +85,10 @@ type OperationHandler[I, O any] interface {
 	// canceled.
 	Start(ctx context.Context, input I, options StartOperationOptions) (HandlerStartOperationResult[O], error)
 	// Cancel handles requests to cancel an asynchronous operation.
-	// Cancelation in Nexus is:
+	// Cancellation in Nexus is:
 	//  1. asynchronous - returning from this method only ensures that cancelation is delivered, it may later be
-	//  ignored by the underlying operation implemention.
-	//  2. idempotent - implementors should ignore duplicate cancelations for the same operation.
+	//  ignored by the underlying operation implementation.
+	//  2. idempotent - implementors should ignore duplicate cancelation for the same operation.
```
chaptersix:
```suggestion
	//  2. idempotent - implementors should ignore duplicate cancelations for the same operation.
```

nexus/options.go
```diff
@@ -18,6 +18,10 @@ type StartOperationOptions struct {
 	//
 	// Implement a [CompletionHandler] and expose it as an HTTP handler to handle async completions.
 	CallbackURL string
+	// CallbackToken is a caller-generated token that is used to recreate context when processing the callback.
+	// CallbackToken must be provided if CallbackURL is not empty.
+	// Handlers must Include the CallbackToken when they invoke the callback.
+	CallbackToken []byte
```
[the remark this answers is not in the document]
bergundy: We are giving this one special treatment because we want to abstract headers away from users.
On the caller side, they will provide an object that will be embedded into a token and let the framework handle serialization.
Now that I'm thinking about it, we should add this already here.

bergundy:
Removed myself as a reviewer. Let get back to this when it becomes higher priority.

PR #65, opened 2025-10-21
[DRAFT] Improve error modeling and serialization

mjameswh:
## Description

This is rough attempt at restoring @bergundy work on improving the `HandlerError`, `OperationError` and `Failure` types to support carrying both a message and a cause at the same time.


nexus/failure.go
```diff
@@ -0,0 +1,241 @@
...
+
+var defaultFailureConverter FailureConverter = knownErrorFailureConverter{}
+
+// DefaultFailureConverter returns the SDK's default [FailureConverter] implementation. Translates [HandlerError],
+// [OperationError] and [FailureError] to and from [Failure] objects maintaining their cause chain.
+// Arbitrary errors are translated to a [Failure] object with its Message set to the Error() string, losing the cause
+// chain.
+// [Failure] instances are converted to [FailureError] to allow access to the full failure metadata and details if
+// available.
+func DefaultFailureConverter() FailureConverter {
```
pdoerner: Does it make sense to allow this to be composed with custom failure converters?

nexus/api.go
```diff
@@ -4,143 +4,94 @@
...
-// [OperationStateFailed] and the given err as the cause.
-//
-// Deprecated: Use [NewOperationFailedError] or construct an [OperationError] directly instead.
-func NewFailedOperationError(err error) *OperationError {
-	return &OperationError{
-		State: OperationStateFailed,
-		Cause: err,
-	}
+	// Set if this error is constructed from a failure object.
+	OriginalFailure *Failure
```
pdoerner: Do we have any constructor helpers for setting this? I don't see it getting set anywhere but maybe I missed it.

nexus/api.go
```diff
@@ -251,38 +208,44 @@ func (e *HandlerError) Retryable() bool {
 	}
 }
 
+func (e *HandlerError) retryBehaviorAsOptionalBool() *bool {
```
pdoerner: Why optional? I would default to non-retryable to make it simpler.

nexus/api.go
```diff
@@ -251,38 +208,44 @@ func (e *HandlerError) Retryable() bool {
...
-	// Token for the operation.
-	Token string `json:"token"`
-	// State of the operation.
-	State OperationState `json:"state"`
-}
-
 // OperationState represents the variable states of an operation.
 type OperationState string
 
+// FIXME(JWH): Should we remove running and succeeded? They are no longer pertinent as public APIs.
```
pdoerner: AFAICT we don't need running. Succeeded might still be useful for completion calls that want to differentiate from failed.

nexus/api.go
```diff
@@ -4,143 +4,94 @@
...
-func (e *FailureError) Error() string {
-	return e.Failure.Message
-}
-
 // OperationError represents "failed" and "canceled" operation results.
 type OperationError struct {
+	// Error message.
+	Message string
+	// Stack trace which may be set if this error was generated by a language that supports it.
+	StackTrace string
```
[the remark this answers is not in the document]
mjameswh: I don't think we should worry about that at this level. If that comes up as an issue, that would have to be handled at the transport level.

mjameswh:
Superseded by other PRs.

PR #69, opened 2025-12-05
Improve error model and serialization

pdoerner:
- 💥 **BREAKING CHANGE** 💥  Removes the experimental `FailureConverter` interface.
- Enhances `HandlerError`, `OperationError`, and `FailureError` to support separate `Cause` and `Message` fields as well as `StackTrace`.
- Deprecates `HandlerErrorf` in favor of `NewHandlerErrorf`.
- Deprecates `OperationErrorf`, `OperationCanceledErrorf`, `OperationFailedErrorf`, `NewOperationFailedError`, `NewOperationCanceledError` in favor of `NewOperationErrorf`, `NewOperationFailedErrorf` and `NewOperationCanceledErrorf`.

Revives the work from https://github.com/nexus-rpc/sdk-go/pull/56 and https://github.com/nexus-rpc/sdk-go/pull/65


nexus/failure.go
```diff
@@ -0,0 +1,239 @@
...
+		var details []byte
+		details, err := json.Marshal(data)
+		if err != nil {
+			return Failure{}, err
+		}
+		f := Failure{
+			Message:    typedErr.Message,
+			StackTrace: typedErr.StackTrace,
+			Metadata: map[string]string{
+				"type": "nexus.HandlerError",
```
VegetarianOrc: I'd consider having this value (`"nexus.HandlerError"`) as a constant or similar.

nexus/failure.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	// NOTE: not using errors.Unwrap here we are intentionally only supporting unwrapping known errors.
+	switch typedErr := err.(type) {
+	case *FailureError:
+		f := typedErr.Failure
+		// Convert the erorr cause if set and failure cause is not already set.
+		if typedErr.Cause != nil && f.Cause == nil {
+			c, err := e.ErrorToFailure(typedErr.Cause)
+			if err != nil {
+				return Failure{}, err
```
VegetarianOrc: Should we consider returning `*Failure` to allow `return nil, err`?

nexus/failure.go
```diff
@@ -0,0 +1,239 @@
...
+	return e.Cause
+}
+
+// FailureConverter is used by the framework to transform [error] instances to and from [Failure] instances.
+// To customize conversion logic, implement this interface and provide your implementation to framework methods such as
+// [NewHTTPClient] and [NewHTTPHandler].
+// By default the SDK translates [HandlerError], [OperationError] and [FailureError] to and from [Failure]
+// objects maintaining their cause chain.
+// Arbitrary errors are translated to a [Failure] object with its Message set to the Error() string, losing the cause
+// chain.
```
bergundy: Might as well mark it as experimental since it's not exposed to Temporal applications yet.

```suggestion
// chain.
//
// NOTE: Experimental
```

bergundy (approved):

bergundy:
Just please don't merge this until we've validated the changes.
I would consider removing the failure converter out of this repo entirely FWIW since it's only used where the protocol is implemented.

nexus/failure.go
```diff
@@ -0,0 +1,240 @@
...
+	Cause *Failure `json:"cause,omitempty"`
+}
+
+// An error that directly represents a wire representation of [Failure].
+// The SDK will convert to this error by default unless the [FailureConverter] instance is customized.
+type FailureError struct {
+	// The underlying Failure object this error represents.
+	Failure Failure
+	// An optional nested cause.
+	Cause error
```
Quinn-With-Two-Ns: `Failure` also has a `Cause`, do we need two different causes? What scenario would both be set?

bergundy: This is the deserialized cause.

nexus/api.go
```diff
@@ -26,88 +25,88 @@ const (
...
+		// Also setting Cause for compatibility with the old implementation.
 		Cause: errors.New(message),
 	}
 }
 
-// OperationFailedErrorf creates an [OperationError] with state set to [OperationStateFailed], using [fmt.Errorf] to
-// construct the cause.
+// OperationFailedErrorf creates an [OperationError] with state set to [OperationStateFailed], using [fmt.Sprintf] to
+// construct the message.
 func OperationFailedErrorf(format string, args ...any) *OperationError {
```
Quinn-With-Two-Ns: This breaks error chaining no? if `args` was an error this used to preserve the error chain and now no longer does.

Quinn-With-Two-Ns: Same goes for `HandlerErrorf`

nexus/api.go
```diff
@@ -26,88 +25,88 @@ const (
...
+	// Error message.
+	Message string
+	// Stack trace which may be set if this error was generated by a language that supports it.
+	StackTrace string
 	// State of the operation. Only [OperationStateFailed] and [OperationStateCanceled] are valid.
 	State OperationState
 	// The underlying cause for this error.
 	Cause error
+	// Set if this error is constructed from a failure object.
+	OriginalFailure *Failure
```
Quinn-With-Two-Ns: I thought we didn't need this anymore since since we are using application failures?

nexus/api.go
bergundy: That was accidentally committed. I meant to revert back to the old implementations. Looks like I missed a couple.

nexus/api.go
bergundy: Still required for the server to rehydrate the original failure.

nexus/errors.go
```diff
@@ -0,0 +1,298 @@
+package nexus
+
+import (
+	"errors"
+	"fmt"
+)
+
+// An error that directly represents a wire representation of [Failure].
+// The SDK will convert to this error by default unless the [FailureConverter] instance is customized.
```
Quinn-With-Two-Ns: `[FailureConverter]` no longer exists in the Nexus SDK right?

bergundy: Good catch!

nexus/errors.go
```diff
@@ -0,0 +1,297 @@
...
+	return &OperationError{
+		State: OperationStateFailed,
+		Cause: errors.New(message),
+	}
+}
+
+// OperationFailedErrorf creates an [OperationError] with state set to [OperationStateFailed], using [fmt.Errorf] to
+// construct the cause.
+//
+// Deprecated: Use NewOperationFailedErrorf instead.
```
Quinn-With-Two-Ns, line 54: If we aren't synchronizing the release of the Nexus SDK with the Temporal Go SDK may not want to deprecate these until the Go SDK that supports causless operation errors is released

Quinn-With-Two-Ns (approved):

PR #70, opened 2026-03-09
Mark some APIs experimental we aren't ready to commit to.

Quinn-With-Two-Ns:
Mark Handler, Serializer and other related concepts as experimental. These APIs may be removed with the expectation that a transport layer implement them rather than the SDK.

See also: https://github.com/nexus-rpc/sdk-python/pull/46


bergundy (approved):
Thanks!

